Issue #33, opened 2018-09-20
Add page-layout, page-type attributes to command adocs


lavenal:
To help with the new server command doc page in the openliberty.io site, the following updates to the asciidoc are required:
- add 
      page-layout: server-command 
   to the following files:
      runnablejarfiles.adoc, server-commands.adoc, server-create.adoc, server-debug.adoc, server-dump.adoc, server-exitcodes.adoc, server-help.adoc, server-javadump.adoc, server-list.adoc, server-package.adoc, server-pause.adoc, server-resume.adoc, server-run.adoc, server-serverprocess.adoc, server-start.adoc, server-stop.adoc, and server-version.adoc

   The page-layout attribute is the name of the layout used to format the doc.
- add 
     page-type: overview
   to server-commands.adoc
- add 
     page-type: command 
   to server-create.adoc, server-debug.adoc, server-dump.adoc,  server-help.adoc, server-javadump.adoc, server-list.adoc, server-package.adoc, server-pause.adoc, server-resume.adoc, server-run.adoc, server-start.adoc, server-stop.adoc, and server-version.adoc 

- add
     page-type: reference
   to runnablejarfiles.adoc, server-exitcodes.adoc, and server-serverprocess.adoc

The page-type attribute is to use for grouping. There are three types in the server command. They are command (that would appear in the table of content), overview (which is an overview doc that appears in the table of content), and reference (that appears in the doc content but would not appear in the table of content).

In addition, the following fixes have to be made:
- server-debug.adoc should have a title of "server debug command" not "server package command"
- server-resume.adoc: remove the following 2 lines

   -yserver --target=MDBApplication#MDBModule.jar#Bean
    -----

lauracowen:
@lavenal @kinueng @mbroz2 Do you know what is meant to be happening with this issue? It looks like work stopped on it and maybe it should be reassigned?

lauracowen:
Have pinged @lavenal on Slack to ask when this work is planned or who this should be re-assigned to.

lavenal:
The work was done back in September with the PR referenced by the issue.

PR #34, opened 2018-09-21
Add page attr to asciidoc

lavenal:
link to issue #33


bsbyrd1:
It looks like two new attributes were added to reflect the fact that these are server command reference docs. What difference should I notice as a result of these attributes being added? Should I be able to see how this update makes the files render on the staging server?

lavenal:
Thanks Brandy. I'm waiting for the code review approval at my end. Once it is done and merged into development branch, you should be able to see the new webpage on the staging website. I'll let you know.

Issue #36, opened 2018-09-26
REST clients (MP06)


bsbyrd1:
- What is it (relative to a REST service), what makes it RESTful? 

- Why and where do you use it? Can a service be a client and a service? alternative ways of doing it?

- Consuming from your own versus third party service - any differences? (eg all our different Java client guides??) 

- Describe CRUD - GET, POST, UPDATE, DELETE

bsbyrd1:
Understanding REST clients 
RESTful services (https://www-03preprod.ibm.com/support/knowledgecenter/SSLVY3_9.7.0/com.ibm.esoatoolkit.doc/topics/c_esoatoolkit_restfulservices.html)

RESTful manifesto (http://roy.gbiv.com/pubs/dissertation/top.htm)
	MP REST client
	JAX-RS-based content is what makes REST specific to Open Liberty

REST resource: http://en.wikipedia.org/wiki/Representational 

Examples that aren’t super detailed; high-level code

Diagrams that describe REST graphically

Good example format: https://openliberty.io/blog/2018/03/22/distributed-in-memory-session-caching.html

lauracowen:
I've moved this back to In progress because I think it needs more work. I've explained more by email to @dmuelle and @andymc12 

I've drafted an outline of the content I think the topic should include: https://ibm.box.com/s/51xt41yg8t8arbf73nhgevnxzht14ppw - subject to input from @andymc12 (or anyone else with more expertise than me in the area of REST clients).

dmuelle:
working draft at https://ibm.ent.box.com/notes/441146861098

dmuelle:
working draft at https://github.com/OpenLiberty/docs/blob/draft/concept/rest_clietns.adoc

dmuelle:
@lauracowen I just checked back w/ Alexandra- she is working on the diagrams today and said she will keep me posted. Once we have that it's ready to publish. When the JSON doc (#35)  and sync/async (#268) docs are published (both are nearing completion) I can go back and add links to them from this doc.

dmuelle:
I've added Alexandra's diagram to the draft site

https://draft-openlibertyio.mybluemix.net/docs/ref/general/#rest_clients.html

I think this is ready to publish, pending final review

dmuelle:
This doc is now live at https://openlibertydev.mybluemix.net/docs/ref/general/#rest_clients.html

lauracowen:
[...]

PR #74, opened 2018-10-18
Minor updates to REST Client doc

andymc12:
- Remove security as a disadvantage for using REST.
- Mention WebSockets as another alternative to REST (in addition to SOAP)
- Remove pseudo-code names from the operations section


andymc12:
#36

PR #97, opened 2019-01-14
Update rol_Java11.adoc

bsbyrd1:
Updating short description.

### DIFF
--- a/ref/rol_Java11.adoc
+++ b/ref/rol_Java11.adoc
@@ -14,19 +14,20 @@
 // Choose a title that a developer would search for, given the subject of the article.
 // PAs of Liberty 18.0.0.4, you can use any version or distribution of Java SE 11.
 
-Significant changes in the Java SE platform occurred between Java SE 8 and Java SE 11. So, zero-migration is not guaranteed.
+You can use any version or distribution of Java SE 11. However, significant changes in the Java SE platform have occurred. Therefore, zero-migration is not guaranteed.
 
 
 == What changed from Java SE 8?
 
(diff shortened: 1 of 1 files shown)


bsbyrd1:
@NottyCode How would you modify the opening paragraph. Be sure to make your suggestions directly to the topic, and I'll have a look!

NottyCode:
@bsbyrd1 For the introduction it says zero migration doesn't apply, but it doesn't provide any limitations on the scope. It could be misinterpreted that there is no longer any guarantee of Zero Migration moving to a fixpack that supports Java 11 from one that doesn't even if Java 11 isn't actually used. I think we need to be care about saying this. I believe we should explicitly state that migrating from Java SE 8 to Java SE 11 may require application migration and if that affects the application there is nothing Liberty can do to prevent it. 

The "zero migration" has always had a limitation that we may not be able to isolate from changes in Java SE platform and I'd like to use this as a chance to say "we can't isolate you" without accidentally implying that all bets are off for zero migration. This also points to a need to document zero migration for Open Liberty.

On TLS 1.3 there are some issues in the Java 11 TLS 1.3 implementation, but we are working to get them fixed. We may have to document some limitations based on Java SE 11 bugs, but I don't expect TLS 1.3 to be disabled as document.

bsbyrd1:
I'm bringing @aguibert into this discussion, since he is contributing to the Java 11 support doc updates for Liberty. Andy, see Alasdair's comment above.

bsbyrd1:
The current topic will become a topic titled, "Java versions supported in Liberty"

PR #130, opened 2019-02-13
Add general folder

ellenwyllie:


kinueng:
This file movement is to align with the URL for these `adoc` files on openliberty.io.

PR #142, opened 2019-03-07
Update microprofile-concurrency.adoc

Rwalls1:


Rwalls1:
@bsbyrd1 I did acrolinx before creating the PR and that is not the score that it received, I will investigate the discrepancy, and I will also address the opening paragraph

PR #147, opened 2019-03-12
Update col_JSONp.adoc

bsbyrd1:
Added a few lines about when to use JSON-B and JSON-B. But it's still written in a way that suggests that JSONB is the better choice. 

@NottyCode  @aguibert Hi guys. Since we support both, can either of you contribute a scenario when JSON-P is the more appropriate choice?

### DIFF
--- a/concepts/col_JSONp.adoc
+++ b/concepts/col_JSONp.adoc
@@ -14,9 +14,11 @@ Combined with other technologies, like JAX-RS, the conversion between JSON and J
 
 == When to use JSON-B and JSON-P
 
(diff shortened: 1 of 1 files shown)


NottyCode:
My understanding is that the big thing that JSON-P has going for it over JSON-B (which could be a possible negative) is that it doesn't lose data. If my JSON-B class has fields "id", "name", "phoneNumber" and the incoming data also has "country" then you lose the country data because you can't fit it into the JSON-B class. This is data loss. Depending on what your code is doing this might be an issue. Of course this could also be a security issue if someone can basically send a massive JSON and the app tries to store it all in memory, although that might also be an issue even with the processing of it for JSON-B.

bsbyrd1:
It seems to be like you said -- it depends on what the code is doing. But it sounds like JSON-P takes the data that it's given, rather than losing what can't fit into the class, which is what happens when using JSON-B. If so, then I'll make that update to the section about JSON-B and JSON-P.

bsbyrd1:
@NottyCode Take a look at the most recent commit, 3cfea7a, and see if that helps to clarify. Thanks!!

PR #198, opened 2019-04-16
Update Java SE support page to mention that Java 12 is now supported

aguibert:


aguibert:
cancelling this PR in favor of #199, which is targeted at the `develop` branch (this PR could not be cleanly re-targeted to that branch)

PR #243, opened 2019-05-06
update-REST-docs

dmuelle:
Fix links and PR issues


dmuelle:
Also addresses OL issue #36

PR #288, opened 2019-05-23
Update to 19.0.0.5 generated doc

NottyCode:


lauracowen:
@NottyCode What updates does this relate to?

NottyCode:
@lauracowen this updates the draft KC so it has the 19.0.0.5 docs in the right format for the draft work we have for inserting examples into the feature doc.

PR #361, opened 2019-06-20
Style edit

lauracowen:
Can you see what you think? Check I've not changed the accuracy by removing some of the capitalisation. If you're happy, I'll do the same for the other versions of the topic.

### DIFF
--- a/ref/feature/jdbc-4.0.adoc
+++ b/ref/feature/jdbc-4.0.adoc
@@ -9,13 +9,13 @@ include::./gen/jdbc-4.0.gendoc[tag=enable]
 
 == Examples
 
-=== Simple DataSource example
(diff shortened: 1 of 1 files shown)


ref/feature/jdbc-4.0.adoc
```diff
@@ -29,7 +29,7 @@ The example below will create a DataSource that can be injected using the name `
 </dataSource>
 ----
 
-A Servlet or EJB can then get it injected for use:
```
NottyCode: I think this should be Servlet because I think it is a proper noun. It is at least a class name.

lauracowen: ok

PR #488, opened 2019-08-06
Update server-windowsservice.adoc

Charlotte-Holt:


ref/commands/server/server-windowsservice.adoc
```diff
@@ -6,74 +6,88 @@
 //
 // Contributors:
 //     IBM Corporation
+:page-layout: server-command
+:page-type: command
+= server WinService command
```
NottyCode: I don't like this title because it doesn't really make sense to me. This isn't a WinService command, it is a set of 4 related commands that are only valid on windows. I think perhaps the title below might be better:

    = Manage Liberty as a Windows service

ref/commands/server/server-windowsservice.adoc
```diff
@@ -6,74 +6,88 @@
...
 
-= server start command
+Register, unregister, start, and stop the named Open Liberty server as a Windows service.
 
-Starts the named Open Liberty server as a Windows service (Windows only).
+The Windows service name and Windows service display name is `serverName`. After you register, you can find and administer the Open Liberty server as a service by using `regedit` from the Windows Registry program. Entries for the service in the registry are in the following locations:
```
NottyCode: I don't think this makes sense where it is. In the original documentation this text was in the context of registerWinService where it makes sense, since it tells you how you can find the windows service in the windows service control panel view, but it doesn't really apply to the start/stop or unregister option.

ref/commands/server/server-windowsservice.adoc
```diff
@@ -6,74 +6,88 @@
...
+----
+HKEY_LOCAL_MACHINE->SOFTWARE->Wow6432Node->Apache Software Foundation->Procrun 2.0->serverName
+HKEY_LOCAL_MACHINE->SYSTEM->CurrentControlSet->services->serverName
+----
 
-== Syntax
+== Usage examples
 
-Run the command from `path_to_liberty/bin`.
+Register the `myserver` server as a Windows service:
```
NottyCode: I'd like to see the commands listed outside of the usage examples.

ref/commands/server/server-windowsservice.adoc
```diff
@@ -6,74 +6,88 @@
...
+----
+server stopWinService myserver
+----
 
-Start `myserver` server in the background:
+Start or stop the `myserver` server by using the server script on Windows systems:
 
 ----
-server start myserver
+server.bat start myserver
```
NottyCode: I don't really know why we are using these since they don't relate in any way to the windows service.

ref/commands/server/server-windowsservice.adoc
```diff
@@ -6,74 +6,88 @@
...
 ----
 
+serverName::
+The name of the server to run the command against. If no server is specified, a server called `defaultServer` is automatically created to run the command against.
 
 == See also
 
-link:server-stop.html[server stop]
-link:server-run.html[server run]
+* command:server-start[]
```
NottyCode: I don't think I'd link to these commands. I'm not sure what purpose it serves.

NottyCode: does this link work?

ref/commands/server/server-windowsservice.adoc
```diff
@@ -6,74 +6,88 @@
...
+* command:server-run[]
+* command:server-stop[]
 
 //See also:
```
NottyCode: should these comments be removed?

ref/commands/server/server-windowsservice.adoc
```diff
@@ -6,74 +6,88 @@
...
 
 //* Equivalent command for Maven.
 //* Equivalent command for Gradle.
-//* Start a Liberty server in the foreground.
-//* Stop a Liberty server.
 //* Create a Liberty server.
 //* Start a Liberty server in debug mode.
 
+== Exit codes
```
NottyCode: I don't think these make sense in this document.

PR #624, opened 2019-09-13
Mp config

ManasiGandhi:


ManasiGandhi:
I am trying to merge the MP config doc with help from @Charlotte-Holt but multiple files are showing up when making a pull request.

lauracowen:
Hmm, sorry. Manasi's change to the MP Config topic looks good, but I just realised there are still all the other 79 topics in this PR. So I won't merge until you've talked with @steven1046 about what the problem is.

PR #753, opened 2019-10-25
Create doc_issue template

chirp1:


chirp1:
@cbridgha Hi Chuck, would you be able to approve so that I can merge the pull request? Also, did I put the template in the right directory so that when authors open a new issue, they get the template?

ISSUE_TEMPLATE/doc_issue.md
```diff
@@ -0,0 +1,34 @@
...
+- Yes ____
+  - Specify the release:____
+- No  ____
+
+### Does the documentation apply to all operating systems? 
+- Yes __  
+- No  __ 
+  - Specify operating systems: ______
+
+### Specify what you want documented (required):
```
NottyCode: shouldn't this be the first thing we ask?

ISSUE_TEMPLATE/doc_issue.md
```diff
@@ -0,0 +1,34 @@
...
+
+#### To create a topic:
+- Specify a link to the section in the navigation where the topic should go:
+- Specify a first draft of the topic that you want added:
+#### To update existing topics:
+- Specify links to the topics that are affected:
+- Include a copy of the current text and the exact text to which it will change. For example: Change ABC to XYZ:
+
+### Author approvals:
+- If you are verifying steps in a task, complete the following items:
```
NottyCode: I don't think the open liberty docs have tasks so I don't think this is needed.

ISSUE_TEMPLATE/doc_issue.md
```diff
@@ -0,0 +1,34 @@
...
+  - Specify the issue for the code bug ____.
+  - Add the Doc bug label.
+- No __
+
+### Does the documentation apply to a particular release?
+- Yes ____
+  - Specify the release:____
+- No  ____
+
+### Does the documentation apply to all operating systems?
```
NottyCode: I wouldn't ask this question. Given the move to containers everything is increasingly linux, so this seems strange. We also don't have a documented list of supporting operating systems

ISSUE_TEMPLATE/doc_issue.md
```diff
@@ -0,0 +1,34 @@
+### Does the doc defect  apply to a code bug?
```
NottyCode: I don't really get this point of this.

chirp1: We ask that question so that we can coordinate the publishing of the doc update with the code update. Would you like the doc and code update to go out together?

ISSUE_TEMPLATE/doc_issue.md
chirp1: We have some issues that are labeled as tasks https://github.com/OpenLiberty/docs/issues?q=is%3Aopen+is%3Aissue+label%3A%22common+task%22   We could end up with a few tasks outside of guides. So, I think it would be good to keep this option in the template. If the issue doesn't deal with a task, the author can skip over this text. Thoughts?

ISSUE_TEMPLATE/doc_issue.md
```diff
@@ -0,0 +1,34 @@
...
+### Does the documentation apply to all operating systems? 
+- Yes __  
+- No  __ 
+  - Specify operating systems: ______
+
+### Specify what you want documented (required):
+Complete the following information:
+
+#### To create a topic:
+- Specify a link to the section in the navigation where the topic should go:
```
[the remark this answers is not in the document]
chirp1: OK I removed that line.

ISSUE_TEMPLATE/doc_issue.md
chirp1: I removed the operating system information.

ISSUE_TEMPLATE/doc_issue.md
chirp1: We ask whether the issue is for a bug and to what release it belongs first so that the author is more likely to provide this information. We use a similar strategy for other templates, and it works.

NottyCode: The template doesn't ask what release it belongs to and it doesn't ask if it is an bug in the documentation either.

ISSUE_TEMPLATE/doc_issue.md
NottyCode: I have 2 issues. The first is that I am not convinced we should have tasks the way KC has tasks. However the bigger issue here is that the entire template talks about topics, doesn't talk about the difference between a topic and a task and then all of a sudden the developer is told if you are doing something with a task without defining what that means. To be honest if this is for a technical accuracy review I don't think it is necessary to have different rules for tasks vs other things. Surely the question is about whether the information is technically correct and they shouldn't need to be told different steps in the case that it is a list of steps to follow vs reference vs concept.

NottyCode:
In general a review approvals section should really be a checklist that is filled out by the reviewer. As a result it should be a comment owned by the reviewer so you have some form of audit tail. In addition as a checklist it would be better off as a series of checkboxes that can be filled out.

If the concern is that the reviewer won't know what to do then the question also comes up when does the developer know that a review is even expected. That seems almost a more fundamental issue.

Also don't forget a template isn't forced, it is there to help people raise issues so they can remove things and I suspect if I saw this in a template I'd be tempted to think that either it is something I'm to do prior to raising the issue, or that it isn't relevant and remove it which is another reason for it being a issue comment.

Should we have different templates for doc bugs vs doc additions/enhancements?

Charlotte-Holt:
THIS IS A TEST 
- [ ] Copy the topicref tags in `INTERNAL__wve_helps_kc.ditamap` to `nav_settings.ditamap`
- [ ] Add  `product="wasnd"` and `platform="dist zos"` to help dita files (replace `outputclass="help"` with `product="wasnd" platform="dist zos"`)
- [ ]  Remove `outputclass="build" role="friend" type="topic"` from the related links and move the related links to the bottom of the dita files
- [ ]  Validate changes in ND and z/OS builds

chirp1:
@NottyCode I think that at this point one template is fine for both defects and enhancements. In a previous comment, a few minutes ago, Charlotte and I tested a checklist. It doesn't  tell who updated which part of the checklist. I like having the reviewer add a label and comment. Both show in the comments who did what..... Adding the labels and comments is something that Marika already has us do for Open Liberty. Here is an example: https://github.com/OpenLiberty/open-liberty/issues/9016   You'll notice that I removed the ID approval label just now as the epic was not approved by ID.

Charlotte-Holt:
THIS IS A TEST
- [ ] Copy the topicref tags in `INTERNAL__wve_helps_kc.ditamap` to `nav_settings.ditamap`
- [ ] Add  `product="wasnd"` and `platform="dist zos"` to help dita files (replace `outputclass="help"` with `product="wasnd" platform="dist zos"`)
- [ ]  Remove `outputclass="build" role="friend" type="topic"` from the related links and move the related links to the bottom of the dita files
- [ ]  Validate changes in ND and z/OS builds

ISSUE_TEMPLATE/doc_issue.md
chirp1: I'm not following your comment. I see "Does the doc update apply to a doc bug? " and "Does the documentation apply to a particular release?". So, the template asks if the update is for a doc bug, and tells the author to specify the release.

ISSUE_TEMPLATE/doc_issue.md
chirp1: Well, topics with steps seem to be creeping into the OpenLiberty documentation. So if any topics have steps, they need to be tested..... Good point about tasks versus topics. I found that I had one reference to tasks and so I changed the wording so that I refer to topics, not tasks..... I do need specific rules for steps in a topic. For years, we've had the rule that topics with steps must be tested. Reviewing the steps doesn't suffice. Errors, unclear instructions, and missing missing explanations are caught when the documentation is tested.

chirp1:
@NottyCode Hi Alasdair, See my latest version of the templte. Charlotte, David, and I reviewed it. I've designed the template so that we can use it to both gather information on what to doc and to  approve what is written.

chirp1:
See #826 and #827 for a continuation of the templates.

Issue #607, opened 2019-09-11
ID: LG-235 mpHealth-2.1


fmhwong:
Epic: https://github.com/openliberty/open-liberty/issues/8655

fmhwong:
`mpHealth-2.1` will **not** be included in 9.0.0.11.

dmuelle:
@fmhwong @pgunapal the updated doc is now on the OL draft site- Are there any other updates needed for this doc?

https://draft-openlibertyio.mybluemix.net/docs/ref/general/#health-check-microservices.html

pgunapal:
@dmuelle The OL draft doc looks good. It has all the required updates.

dmuelle:
doc is published at https://www.openliberty.io/docs/ref/general/#health-check-microservices.html

PR #805, opened 2019-11-19
Mueller 607 updates

dmuelle:


dmuelle:
Updates for MP Health via #607

ref/general/health-check-microservices.adoc
```diff
@@ -25,15 +25,15 @@ A self-check can assess anything that the service needs:
...
-- `/health`: aggregates the responses from `/health/live` and `/health/ready`. This endpoint corresponds to the deprecated `@Health` annotation and is only available to provide compatibility with MicroProfile 1.0. Developers who are using MicroProfile 2.0 or higher should use the `/health/ready` or `/health/live` endpoints instead.
+- `/health/live`: returns the liveness of a service, or whether it encountered a bug or deadlock. If this check fails, the service is not running and can be stopped. This endpoint corresponds to the Kubernetes liveness probe, which automatically restarts the pod if the check fails.
+- `/health`: aggregates the responses from `/health/live` and `/health/ready`. This endpoint corresponds to the deprecated `@Health` annotation and is only available to provide compatibility with MicroProfile 1.0. Developers who are using MicroProfile 2.0 or higher can use the `/health/ready` or `/health/live` endpoints instead.
```
lauracowen: "should" is the right word here. We don't want developers who are able to use MP2.0 using the old endpoints. We can't say "must" because that would raise the question of "or else what?". Alternatively rephrase the sentence to "If you are using MP 2 or higher, use the ....endpoints instead." (actually I prefer that)

ref/general/health-check-microservices.adoc
```diff
@@ -25,15 +25,15 @@ A self-check can assess anything that the service needs:
...
+For example, in a microservices-based banking application, you might implement health checks on the login service, the balance transfer service, and the bill pay service. If a health check on the balance transfer service detects a bug or deadlock, it reports as DOWN. In this case, if the `/health/live` endpoint is configured in a Kubernetes liveness probe, the probe restarts the service automatically. This feature saves the user from encountering downtime or an error and preserves the functions of the other services in the application.
 
 == MicroProfile Health provided endpoints and annotations
 
-MicroProfile Health provides three endpoints:
+MicroProfile Health provides the following three endpoints:
```
lauracowen: I think this makes the sentence unnecessarily wordy. Having a colon at the end signals the start of a list of the three endpoints so I don't think this adds anything even with a screenreader.

ref/general/health-check-microservices.adoc
```diff
@@ -11,11 +11,11 @@
...
-= Enabling health checking of microservices
+= Enabling microservice health checks
 
-Monitoring the health of the microservices that make up your application is crucial to providing your users with reliable functionality. A _health check_ is a special REST API implementation that you can use to validate the status of a microservice and its dependencies. link:https://github.com/eclipse/microprofile-health[MicroProfile Health] enables services in an application to self-check their health and then publishes the overall health status to a defined endpoint.
+Monitoring the health of the microservices that make up your application is crucial to providing your users with reliable performance. A _health check_ is a special REST API implementation that you can use to validate the status of a microservice and its dependencies. link:https://github.com/eclipse/microprofile-health[MicroProfile Health] enables services in an application to self-check their health and then publishes the overall health status to a defined endpoint.
 
-A self-check can assess anything that the service needs:
+A self-check can assess anything that the service needs, including the following factors:
```
lauracowen: I disagree. Having a colon after "...needs:" signals the start of a list. "factors" doesn't mean anything and it's just adding padding to the sentence.

ref/general/health-check-microservices.adoc
```diff
@@ -25,15 +25,15 @@ A self-check can assess anything that the service needs:
...
 
-For example, in a microservices-based banking application, health checks might be implemented on the login service, the balance transfer service, and the bill pay service. If a health check on the balance transfer service detects a bug or deadlock, it reports as DOWN. In this case, if the `/health/live` endpoint is configured in a Kubernetes liveness probe, the probe restarts the service automatically, saving the user from encountering downtime or an error and preserving the functions of the other services in the application.
+For example, in a microservices-based banking application, you might implement health checks on the login service, the balance transfer service, and the bill pay service. If a health check on the balance transfer service detects a bug or deadlock, it reports as DOWN. In this case, if the `/health/live` endpoint is configured in a Kubernetes liveness probe, the probe restarts the service automatically. This feature saves the user from encountering downtime or an error and preserves the functions of the other services in the application.
```
lauracowen: I agree that breaking the last sentence into two shorter ones makes sense but I don't think "This feature saves..." is any more informative than just saying "This saves..." (which is I'm guess the reason for saying "feature". The word "feature" is overloaded and is just used here as a filler. Instead, could reword the second sentence to start "Restarting the service automatically saves the user..."?

PR #826, opened 2019-12-03
Doc bug template

chirp1:


Charlotte-Holt:
We originally created these templates to address RHR doc requirements, but I think now the labels that we have are sufficient for that purpose. 

Closing these, and can reopen them if we decide we need similar templates in the future.

PR #925, opened 2020-01-17
Add extended description paragraph

lauracowen:
~Added `description` directory and a `sociallogin.description` file in it to add an extended description to the Social Login generated feature doc.~

@NottyCode I've restructured the examples and description directories to group the new files by feature name. What do you think?

If you're happy with the approach, we should maybe change the tag name of the summary sentence to something like `sentence` or `summary` instead of it being called `description` too?

@emilytee See the audit-related files. See also the runtime PR: https://github.com/OpenLiberty/open-liberty/pull/10527


lauracowen:
> The extended description for this one is already enroute on [OpenLiberty/open-liberty#10373](https://github.com/OpenLiberty/open-liberty/pull/10373) so not needed here but it's a good idea and we should put it in the next feature we add documentation too.

@brutif After discussion with Alasdair, I've submitted a PR to change your properties edit back to a single sentence (but the much better sentence you put in there) and moved the rest of it into the manual docs. See https://github.com/OpenLiberty/open-liberty/pull/10527 for change currently available to review.

brutif:
@lauracowen that's fine, thanks for letting me know

ref/feature/audit/description.adoc
```diff
@@ -0,0 +1,31 @@
+The Audit feature introduces an infrastructure that serves two purposes: \n\n\
```
NottyCode: the line endings should be removed I think. They cause a properties file to insert line breaks and disable the physical line break meaning the next line is a properties.

ref/feature/audit/description.adoc
```diff
@@ -0,0 +1,31 @@
...
+ - Confirming the effectiveness and integrity of the existing configuration \n\
+ - Identifying areas where improvement to the configuration may be needed  \n\n\
+
+The Audit feature can capture a range of auditable events, including \n\n\
+events related to authentication, authorization, and logout.
+//  - Basic authentication \n\
```
NottyCode: Is the // intended? If it comments out the bullet I'd rather just remove the bullets.

ref/feature/socialLogin/socialLogin copy.examples
```diff
@@ -0,0 +1,89 @@
+== About this feature
```
NottyCode: Is this an accidental checkin?

lauracowen: Yeah :( sorry, I thought I was being careful not to do that but it clearly slipped through.

ref/feature/audit/description.adoc
lauracowen: I left them in in case you wanted more detail adding back in but I'll remove them now.

ref/feature/audit/description.adoc
lauracowen: Oops, yes, will remove.

Issue #893, opened 2020-01-07
Authentication in Open Liberty


Charlotte-Holt:
From LC: This is a new concept topic about authentication in Open Liberty. Authentication verifies who the user is. User ID and password (user registry - link to separate topic),  SSO (LTPA, OpenID Connect, SAML, JWT, Social login, SPNEGO - link to separate topics because only one or two of them will be relevant to any given reader so they can ignore the ones they're not interested in). Users of apps running on OL can use their social media accounts (using OAuth and Open ID standards) to log into the apps. If multiple options are presented with a form to select which (give example screenshot to illustrate?). Out of the box, apps can be configured to enable users to use Facebook, Twitter, GitHub, LinkedIn, and Google, but others can be added. Once the user has authenticated, show how an app can find out who the person is.

The authentication topic in the KC provides a bit of an intro but it needs more. First introduce the concept of authentication on Open Liberty needs to be introduced first Introduce what JAAS (Java Authentication and Authorization Service) is and how it is used in authentication processes.

A diagram is needed but I don't know whether the one in this topic is a good one to use or not - probably needs updating for the context of microservices - ideally using one of our existing scenarios such as the system properties or the music store. 
- https://www.ibm.com/support/knowledgecenter/SSD28V_liberty/com.ibm.websphere.wlp.core.doc/ae/cwlp_authentication.html 
- https://www.ibm.com/support/knowledgecenter/SSD28V_liberty/com.ibm.websphere.wlp.core.doc/ae/cwlp_auth_app_client.html 
- https://ibm.box.com/s/b9d5fililsqtwsb3yu8en3kqwet0ocsc
- https://ibm.box.com/s/thvnli5i160y63gk24ybs3ekvuj1p4h1 - presentation by Jim Mulvey
- SPNEGO config examples w/ #954 
- JWT config examples w/ #636

dmuelle:
@lauracowen I can't access either of the box notes above- the second appears to be expired and I don't have permission to the first- if another link is available can you share with me? The KC resources are helpful but I need to know what is specific to OL. I'm going to work through the `Securing a web application` guide, which should help as well.

lauracowen:
Oops, sorry. The first was to the Intro to Security MP4 video and the second is the slides to that video. You can find them both in here: https://ibm.box.com/s/wokyhx7j1yr8883ehpdcnj2irzvao1xx

lauracowen:
[...]

PR #981, opened 2020-01-30
edits per LC review

dmuelle:
addresses suggestions from comment in #893


dmuelle:
@lauracowen - actually that statement needed a little tweaking after all- it's corrected now in a new commit

PR #983, opened 2020-01-31
707 standalone edits

dmuelle:


dmuelle:
@lauracowen - moved the Operators section to last and reduced the diagram size by 75%. Note that the previous size matched what we had on the site. I assume the back end will constrain/expand it as needed since there wasn't anything in the image statement to do so...

PR #990, opened 2020-01-31
Updates to the server config overview topic

Charlotte-Holt:
@NottyCode, these are the updates to the server configuration overview that I mentioned. Please review the changes and let me know if you have any questions.


ref/config/serverConfiguration.adoc
```diff
@@ -3,13 +3,7 @@
...
 :page-type: overview
 
-The {projectName} server config is made up of one mandatory file, the
-`server.xml` file, and a set of optional additional files. The `server.xml` is
-the only mandatory config file.  The only requirements for the `server.xml`
-file are that it must be well formed XML and the root element is `server`. When
-processing the `server.xml` file, any elements or attributes that are not
-understood are simply ignored. The default `server.xml` might look like the
-following example:
+The {projectName} server configuration, or server config, is made up of one mandatory file, the `server.xml` file, and a set of optional files. The `server.xml` file must be well-formed XML and the root element must be `server`. When the `server.xml` file is processed, any elements or attributes that are not understood are ignored. The default `server.xml` file might look like the following example:
```
NottyCode: I think we should stick with server config. I think config is a widely understood abbreviation for configuration so I don't think we need to clarify.

ref/config/serverConfiguration.adoc
```diff
@@ -24,171 +18,134 @@ following example:
...
-expand WAR files when they are deployed.
+This `server.xml` file configures the server to do the following things:
 
-The `server.xml` is described in more detail <<server-xml,below>>. When
-discussing config for {projectName}, the term "server config" can be used to refer
-to all of the files that make up the server config, or specifically to the
-config held in XML files; it is usually clear in context and, if not, the
-phrase "server XML config" might be used to clarify. The next few sections
-describe each of the config files in the order they are processed.
+* Support the Java Server Pages 2.3 feature
```
NottyCode: I think switching to bullets should only happen if we reference back to the server.xml snippet which asciidoc can do. It can add markers to the xml and cross reference making it clear what is referring to. Otherwise I think sticking with the more compact form is better because I don't think what the xml does is as important as what it looks like and I think using bullets focuses too much on what it does vs what it looks like.

ref/config/serverConfiguration.adoc
```diff
@@ -24,171 +18,134 @@ following example:
...
+* Listen to incoming traffic to `localhost` on port `9080`
+* Automatically expand WAR files when they are deployed
 
-== server.env
-The `server.env` files are optional but, when present, are read by the
-`bin/server` script and specify environment variables primarily used to
-influence the behavior of the `bin/server` script. `server.env` files are read
-from the following locations in order:
+The term _server config_ can be used to refer to all of the files that make up the server configuration, or specifically to the configuration that's in the XML files. If it's not clear in context, you can use the term _server XML config_ to refer to the configuration in the XML files.
```
NottyCode: The original version of this was describing how we use the term server config and server XML config in this document and elsewhere. Generally we use server config unless there is ambiguity. This changes it to be about the reader as a result it doesn't achieve the function it is intended to which is to explain terminology used in this document.

ref/config/serverConfiguration.adoc
```diff
@@ -24,171 +18,134 @@ following example:
...
 
-* `${wlp.install.dir}/etc/`
-* `${wlp.user.dir}/shared/`
-* `${server.config.dir}/`
+== Configuration files
```
NottyCode: It is not clear what value this section provides.

ref/config/serverConfiguration.adoc
```diff
@@ -24,171 +18,134 @@ following example:
...
-  is used which often isn't desired.
-* `WLP_USER_DIR` - indicates the location of the `usr` directory which contains
-  the server config. This can only be set in the `etc/server.env` file because
-  the other locations are relative to the `usr` directory.
-* `WLP_OUTPUT_DIR` - indicates where the server should write files to. By
-  default, the server writes to the directory structure that the config is
-  read from. However, in some secure profiles the server config needs to
-  be read-only and so the server must write files to another location.
+[#server-env]
+=== `server.env`
```
NottyCode: I don't like using `` around the title. It makes the styling of the headings look terrible.

ref/config/serverConfiguration.adoc
```diff
@@ -24,171 +18,134 @@ following example:
...
-`${wlp.install.dir}/etc` and uses that.
+The `server.env` file is in `KEY=value` format, as shown in the following example:
+
+[source,properties]
+----
+JAVA_HOME=/opt/ibm/java
+WLP_USER_DIR=/opt/wlp-usr
+----
+
+Key values must not contain white space. The values are interpreted literally so you don't need escape spaces or other special characters. These files don't support variable substitution.
```
NottyCode: We should probably not say key values must not contain spaces, because it is space escaping it shouldn't contain. Error in prior copy.

ref/config/serverConfiguration.adoc
```diff
@@ -24,171 +18,134 @@ following example:
...
+
+If no `jvm.options` files exist in these locations, then the server script looks for the file in `${wlp.install.dir}/etc`, if such a directory exists.
 
 Common uses of `jvm.options` files include:
 
 * Setting JVM memory limits
-* Enabling Java Agents provided by monitoring products
-* Setting Java System Properties
+* Enabling Java Agents that are provided by monitoring products
+* Setting Java system properties
```
NottyCode: I think System Properties is likely correct since System refers to a class `java.lang.System` and the Properties are accessible via `getProperties`

ref/config/serverConfiguration.adoc
```diff
@@ -24,171 +18,134 @@ following example:
...
 -Dmy.system.prop=This is the value.
 ----
 
-There is no need to escape special characters, such as spaces.
-
-The options are read and provided in order to the JVM. If you provide multiple
-options they are all seen by the JVM (most JVMs use the last option on the
-command line and ignore prior options). Certain options must not be put in the
-`jvm.options` file such as `-jar`.
+You don’t need escape spaces or other special characters. Options are read and provided to the JVM in order. If you provide multiple options, then they are all seen by the JVM. These files do not support variable substitution.
```
NottyCode: I think the change to space or other special characters just begs the question what are the special characters. It makes it sound like spaces are not special characters. I think it is better the original way where spaces is an example of a special character.

ref/config/serverConfiguration.adoc
```diff
@@ -24,171 +18,134 @@ following example:
...
-dropping server-formatted XML files into directories. Within the two
-`configDropins` directories the files are read in alphabetical order.
+The `${server.config.dir}/server.xml` file must be present, but the other files are optional. You can flexibly compose config by dropping server-formatted XML files into directories. Files are read in alphabetical order in each of the two `configDropins` directories.
 
-=== Variable substitution
-You can parameterize server config using variables. When resolving
-variable names the following sources are consulted in increasing order of
-precedence:
+== Variable substitution
+You can use variables to parameterize the server configuration. To resolve variable names, the following sources are consulted:
```
NottyCode: upon reflection I don't think we are resolving variable names so much as resolving variable references to their values.

ref/config/serverConfiguration.adoc
```diff
@@ -24,171 +18,134 @@ following example:
...
 
-* `server.xml` default variable values
-* environment variables
+* `server.xml` file default variable values
+* Environment variables
 * `bootstrap.properties`
 * Java system properties
-* `server.xml` config
+* `server.xml` configuration
```
NottyCode: we use config elsewhere so why is this configuration?

ref/config/serverConfiguration.adoc
```diff
@@ -24,171 +18,134 @@ following example:
...
-=== Variable substitution
-You can parameterize server config using variables. When resolving
-variable names the following sources are consulted in increasing order of
-precedence:
+== Variable substitution
+You can use variables to parameterize the server configuration. To resolve variable names, the following sources are consulted:
 
-* `server.xml` default variable values
-* environment variables
+* `server.xml` file default variable values
```
NottyCode: why is this file necessary whereas elsewhere we are ok just saying `server.xml` config?

ref/config/serverConfiguration.adoc
```diff
@@ -24,171 +18,134 @@ following example:
...
+* Environment variables
 * `bootstrap.properties`
 * Java system properties
-* `server.xml` config
+* `server.xml` configuration
 
-Variables are referenced using `${variableName}` syntax. In server config,
-specify variables using the `variable` element:
+In this list, each source takes precedence over the previous source, so that the `server.xml` configuration takes the highest precedence and the  `server.xml` file default variable values take the lowest precedence. Variables are referenced by using the `${variableName}` syntax. You can specify variables in the server configuration by using the  `variable` element, as shown in the following example:
```
NottyCode: If the prior list were numbered rather than bulleted then the first added sentence would not be necessary.

NottyCode: Given there is an example following I don't know why we need to spell out the variable element in the preceding sentence.

ref/config/serverConfiguration.adoc
```diff
@@ -24,171 +18,134 @@ following example:
...
 
 [source,xml]
 ----
 <variable name="variableName" value="some.value" />
 ----
 
-Default values, specified in server config, are only used if no other value can
-be found. They are specified using the `variable` element and the `defaultValue`
-attribute:
+Default values, which are specified in the `server.xml` file, are used only if no other value is specified. These values are specified by using the `variable` element and the `defaultValue` attribute:
```
NottyCode: similar comment to above.

ref/config/serverConfiguration.adoc
```diff
@@ -24,171 +18,134 @@ following example:
...
 
-Environment variables can be accessed as variables. From 19.0.0.3, they can be
-accessed directly by referencing the environment variable name. If the variable
-cannot be resolved the following transformations on the environment variable
-name is tried:
+Environment variables can be accessed as variables. As of version 19.0.0.3, you can reference the environment variable name directly. If the variable cannot be resolved as specified, the `server.xml` file looks for the following variations on the environment variable name:
 
-1. Replace all non-alpha num characters with _
-2. Change all characters to upper case.
+* Replacing all non-alphanumeric characters with the underscore character (`_`)
```
NottyCode: I don't think using ing words makes it clearer what is going on.

ref/config/serverConfiguration.adoc
```diff
@@ -24,171 +18,134 @@ following example:
...
+* Changing all characters to uppercase
 
-If you enter `${my.env.var}` in `server.xml` it will look for environment
-variables with the following names:
+For example, if you enter `${my.env.var}` in the `server.xml` file, it looks for environment variables with the following names:
 
-1. my.env.var
-2. my_env_var
-3. MY_ENV_VAR
+* my.env.var
```
NottyCode: The numbers were there to show precedence order. Changing to bullets removes this necessary ordering from the options.

ref/config/serverConfiguration.adoc
```diff
@@ -221,37 +175,28 @@ operator can be one of `+`, `-`, `*`, `/` for example:
 <variable name="threeagain" value="${six/two}" />
 ----
 
-There are a number of predefined variables:
+The following predefined variables are supported:
```
NottyCode: I don't think "are supported" is correct. They exist, they are built in. Are supported optionality that doesn't exist.

ref/config/serverConfiguration.adoc
```diff
@@ -221,37 +175,28 @@ operator can be one of `+`, `-`, `*`, `/` for example:
...
 
-* `wlp.install.dir` - the location where the Liberty runtime is installed.
-* `wlp.server.name` - the name of the server.
-* `wlp.user.dir` - the location of the `usr` folder. Defaults to
+* `wlp.install.dir`: the location where the Open Liberty runtime is installed.
```
NottyCode: This won't render well because of the format of the code marked things. Recommend sticking with -

ref/config/serverConfiguration.adoc
```diff
@@ -221,37 +175,28 @@ operator can be one of `+`, `-`, `*`, `/` for example:
...
+* `wlp.user.dir`: the location of the `usr` folder. The default is
   `${wlp.install.dir}/usr`.
-* `shared.app.dir` - the location of shared applications. Defaults to
+* `shared.app.dir`: the location of shared applications. The default is
   `${wlp.user.dir}/shared/apps`.
-* `shared.config.dir` - the location of shared configuration files. Defaults to
+* `shared.config.dir`: the directory that contains the server config. The default is
   `${wlp.user.dir}/shared/config`.
-* `shared.resource.dir` - the location of shared resource files. Defaults to
+* `shared.resource.dir`: the location of shared resource files. The default is
```
NottyCode: shared.config.dir is referenced as a directory over location, but shared.resource.dir is sticking with location?

ref/config/serverConfiguration.adoc
```diff
@@ -221,37 +175,28 @@ operator can be one of `+`, `-`, `*`, `/` for example:
...
-config. Merging works differently for the two. Singleton config
-is used when configuring a single thing (e.g. logging); factory config is
-used when it is valid to configure multiple things, (e.g. an application or a
-data source).
-
-==== Merging singleton config
-
-For singleton config elements the config is merged. If two
-elements exist with different attributes both attributes are used. For example:
+* `server.config.dir`: the directory that server config is stored in.
```
NottyCode: see earlier comment about : and -

ref/config/serverConfiguration.adoc
```diff
@@ -221,37 +175,28 @@ operator can be one of `+`, `-`, `*`, `/` for example:
...
+  The default is `${wlp.user.dir}/servers/${wlp.server.name}`.
+* `server.output.dir`: the directory where the server writes the workarea, logs, and
+  other runtime-generated files. The default is `${server.config.dir}`.
+
+== Configuration merging
```
NottyCode: See prior comments about Config vs Configuration

ref/config/serverConfiguration.adoc
Charlotte-Holt: Updated

ref/config/serverConfiguration.adoc
Charlotte-Holt: Updated

ref/config/serverConfiguration.adoc
Charlotte-Holt: Updated

ref/config/serverConfiguration.adoc
Charlotte-Holt: Updated

ref/config/serverConfiguration.adoc
Charlotte-Holt: Updated

ref/config/serverConfiguration.adoc
Charlotte-Holt: Updated in both places

ref/config/serverConfiguration.adoc
Charlotte-Holt: Updated

ref/config/serverConfiguration.adoc
Charlotte-Holt: Updated

ref/config/serverConfiguration.adoc
Charlotte-Holt: Updated

ref/config/serverConfiguration.adoc
Charlotte-Holt: Updated

ref/config/serverConfiguration.adoc
Charlotte-Holt: Updated in both places

ref/config/serverConfiguration.adoc
Charlotte-Holt: Updated

ref/config/serverConfiguration.adoc
Charlotte-Holt: Updated

ref/config/serverConfiguration.adoc
Charlotte-Holt: Updated

ref/config/serverConfiguration.adoc
Charlotte-Holt: I think that was an error. Updated.

ref/config/serverConfiguration.adoc
Charlotte-Holt: Updated

ref/config/serverConfiguration.adoc
Charlotte-Holt: Updated

ref/config/serverConfiguration.adoc
Charlotte-Holt: Updated both of these. See if the list still reads properly to you with the updates.

ref/config/serverConfiguration.adoc
Charlotte-Holt: Updated

ref/config/serverConfiguration.adoc
```diff
@@ -221,37 +175,28 @@ operator can be one of `+`, `-`, `*`, `/` for example:
...
+Since the configuration can consist of multiple files, it is possible that two files provide the same configuration. In these situations, the server config is merged according to a set of simple rules. In {projectName}, configuration is separated into Singleton and factory configuration, each of which has its own rules for merging. Singleton config is used to configure a single element (for example, logging). Factory config is used to configure multiple entities, such as an entire application or data source.
```
NottyCode: Singleton isn't a proper noun so should be a lower case s.

More inconsistent use of config vs configuration

ref/config/serverConfiguration.adoc
```diff
@@ -221,37 +175,28 @@ operator can be one of `+`, `-`, `*`, `/` for example:
...
+
+=== Merging Singleton configuration
```
NottyCode: singleton not Singleton

ref/config/serverConfiguration.adoc
```diff
@@ -221,37 +175,28 @@ operator can be one of `+`, `-`, `*`, `/` for example:
...
+For Singleton config elements that are specified more than once, the config is merged. If two elements exist with different attributes, both attributes are used. For example:
```
NottyCode: singleton not Singleton

ref/config/serverConfiguration.adoc
```diff
@@ -290,9 +234,7 @@ is treated as:
 </server>
 ----
 
-In some cases, config is provided using child elements that take text. In
-these cases the config is merged by using all of the values specified. The most
-common scenario is configuring features. For example:
+In some cases, config is provided by using child elements that take text. In these cases, the config is merged by using all of the values specified. The most common scenario is configuring features. For example:
```
NottyCode: I'm tripping up over the added commas

ref/config/serverConfiguration.adoc
```diff
@@ -318,18 +260,11 @@ is treated as:
...
-
-Factory config merges use the same rules as singleton config but, because
-it is valid to configure the same element and mean two different logical objects,
-merging doesn't happen just because the element names match. Instead each
-element is assumed to be configuring a distinct object. If the logical object is
-configured by two instances, the `id` attribute must be set on each of them
-to indicate they are the same thing. Variable substitution on an `id` is not
-supported.
+=== Merging factory configuration
+Factory config merges use the same rules as Singleton config except elements are not automatically merged just because the names match. Unlike Singleton config, with factory config it is valid to configure the same element and mean two different logical objects. Therefore, each element is assumed to configure a distinct object. If the logical object is configured by two instances, the `id` attribute must be set on each of them to indicate they are the same thing. Variable substitution on an `id` attribute is not supported.
```
NottyCode: singleton.

it says not merged just because the names matched, but it doesn't specify what a name is. In this case it means the element name, but I think that needs to be clarified other wise it could be inferred that it means the name attribute.

This reads incorrectly vs the actual behaviour:

>If the logical object is configured by two instances, the `id` attribute must be set on each of them to indicate they are the same thing.

You are saying to have two elements configure the same object you need to use the id attribute with the same value, but the way it reads with reference to two instances as expressed it makes it sound like you have to specify the id attribute if you want 2 logical instances, rather than one.

ref/config/serverConfiguration.adoc
```diff
@@ -341,14 +276,9 @@ the context root:
...
-=== Include processing
+== Include processing
 
-In addition to the default locations, additional config files can be
-brought in using the `include` element. When a server config file contains an
-`include` reference to another file, the server processes the contents of the
-referenced file as if they were included inline in place of the `include`
-element. In the following example, the server processes the contents of the
-`other.xml` file before processing the contents of the `other2.xml` file:
+In addition to the default locations, more configuration files can be brought in by using the `include` element. When a server config file contains an include reference to another file, the server processes the contents of the referenced file as if they were included inline in place of the `include` element. In the following example, the server processes the contents of the `other.xml` file before it processes the contents of the `other2.xml` file:
```
NottyCode: I don't like "more configuration files" and configuration vs config is yet again made inconsistent in the docs.

ref/config/serverConfiguration.adoc
```diff
@@ -358,8 +288,7 @@ element. In the following example, the server processes the contents of the
 </server>
 ----
 
-By default, an include file must exist but, if the include file might not be
-present, the `optional` attribute can be set to `true`. For example:
+By default, the specified include file must exist. If the include file might not be present, set the `optional` attribute to `true`, as shown in the following example:
```
NottyCode: I think specified is redundant.

ref/config/serverConfiguration.adoc
```diff
@@ -380,20 +307,13 @@ conflicting config or to replace it (`REPLACE`).
...
 
-=== Config references
-Most configuration in {projectName} is self-contained but it is often useful to
-be able to share config. A common example of this would be the JDBC driver
-config being shared by multiple data sources, or sharing the classloader for
-JDBC driver classes so the classes are visible both to the DataSource and to an
-application. Any factory config element defined as a direct child of the
-`server` element can be referred to.
+== Configuration references
+Most configuration in Open Liberty is self-contained but it can be useful to share config. For example, the JDBC driver config might be shared by multiple data sources. You can refer to any factory config element that is defined as a direct child of the `server` element.
```
NottyCode: Open Liberty should be {projectName}

The rephrasing reduces the utility of the documentation. Before it says "here is something many people need/want to do" and it goes to "something that some people might need, but who knows if it is common". I prefer the original wording because it enhances the utility of the information being conveyed.

ref/config/serverConfiguration.adoc
Charlotte-Holt: Updated

ref/config/serverConfiguration.adoc
Charlotte-Holt: Updated

ref/config/serverConfiguration.adoc
Charlotte-Holt: Updated

ref/config/serverConfiguration.adoc
Charlotte-Holt: Updated

ref/config/serverConfiguration.adoc
Charlotte-Holt: Updated

ref/config/serverConfiguration.adoc
Charlotte-Holt: Updated

ref/config/serverConfiguration.adoc
Charlotte-Holt: Updated

ref/config/serverConfiguration.adoc
Charlotte-Holt: This wasn't changed from the original:
> > If the logical object is configured by two instances, the `id` attribute must be set on each of them to indicate they are the same thing.

Would it be better as:
If a logical object is configured by two elements, the `id` attribute must be set on each element to indicate they are the same thing.

ref/config/serverConfiguration.adoc
Charlotte-Holt: Updated - check that the descriptions are adequate

ref/config/serverConfiguration.adoc
NottyCode: sorry I had missed the original error was mine not yours. What about:

> If a single logical object is configured by two elements, the id attribute must be set on each element to indicate they are the same thing.

ref/config/serverConfiguration.adoc
Charlotte-Holt: Updated again

ref/config/serverConfiguration.adoc
Charlotte-Holt: Updated.

ref/config/serverConfiguration.adoc
Charlotte-Holt: Updated again

ref/config/serverConfiguration.adoc
NottyCode: So I'm still not sure about the descriptions, but I'm not going to hold it up.

PR #993, opened 2020-02-03
Update hardening-intro.adoc

Charlotte-Holt:


ref/general/hardening-intro.adoc
```diff
@@ -13,7 +13,7 @@
 :page-type: general
 = Hardening Open Liberty servers in production
 
-Hardening is a general term that is used to describe the measures that are taken to enhance your system security against potential threats and risks. It is important to harden your Open Liberty server to reduce vulnerabilities and prevent against different types of intrusions that can occur. Security intrusions and vulnerabilities can be broken down into different groups:
+Hardening is a general term that is used to describe the measures that are taken to enhance your system security against potential threats and risks in production. It is important to harden your Open Liberty server to reduce vulnerabilities and prevent against different types of intrusions that can occur. Different types of security intrusions can threaten different parts of your system, and there are best practices that you can employ to make each part less vulnerable. Security intrusions and vulnerabilities can be broken down into different groups:
```
dmuelle: `It is important to harden your Open Liberty server to reduce vulnerabilities and prevent against different types of intrusions that can occur.`

could be more concise-

`It is important to harden your Open Liberty server to reduce vulnerabilities and prevent security intrusions.`

ref/general/hardening-intro.adoc
```diff
@@ -22,49 +22,38 @@ Hardening is a general term that is used to describe the measures that are taken
...
-
-===== Carefully configure file permissions
-When you configure file permissions, ensure that those who want access to the configuration fall into the appropriate permission bit. People who shouldn't have access at all fall into the `other` permission bit that's marked as `0`.
-
-===== Encrypt passwords
-Use encryption for passwords instead of Base64 encoding. You can use the `securityUtility` command, which provides plain text encryption for Open Liberty.
+* *Keep Open Liberty up to date*: To reduce operating system vulnerabilities, it's important to keep your version of Open Liberty up to date. Open Liberty fix packs are compatible with later versions, and the zero migration policy means that you can move to a newer fix pack without changing anything in your server configuration. Ensuring that you have a recent version of Open Liberty positions you for link:/docs/ref/general/#security-vulnerabilities.html[critical security fixes].
```
dmuelle: `Ensuring that you have a recent version of Open Liberty positions you for link:/docs/ref/general/#security-vulnerabilities.html[critical security fixes].`

`Keeping Open Liberty up to date helps ensure that you have the most recent link:/docs/ref/general/#security-vulnerabilities.html[critical security fixes].`

dmuelle: should there be a link to DL the most recent version here?

ref/general/hardening-intro.adoc
```diff
@@ -22,49 +22,38 @@ Hardening is a general term that is used to describe the measures that are taken
...
+* *Use user registries that are safe for production*: Basic and Quick Start Security user registries are useful for development environments, but they are not acceptable for readiness testing, QA, or production use. Instead, use an LDAP user registry, federated repositories, or create a custom user registry. If you use an LDAP user registry, maintain your bind password in a separate `include` file, not in the `server.xml` file. If you use a custom registry, use the `UserRegistry` class to implement the registry, and make sure that the registry is thoroughly reviewed before you use it in production.
+* *Carefully configure file permissions*: When you configure file permissions, ensure that the users who want access to the configuration fall into the appropriate permission bit. Users that are not the owner and are not part of the authorized group fall into the `other` permission bit, which is marked as `0`.
+* *Encrypt passwords*: Use encryption for passwords instead of Base64 encoding. You can use the `securityUtility` command, which provides plain text encryption for Open Liberty. For more information about the `securityUtility command`, go to <link pending>.
```
dmuelle: need to double check with Ajay on the terminology here. In my user registry topic, he recommended using "hash encoding" as opposed to encrypting. They are two separate things. We should clarify with him which is best practice. I need to know as well and then make this section, my doc, and the forthcoming securityUtility doc consistent.

ref/general/hardening-intro.adoc
```diff
@@ -22,49 +22,38 @@ Hardening is a general term that is used to describe the measures that are taken
...
+==== Best practices
 
-===== alsdkfj;asdf
-Configure applications to run on a virtual host to keep admin traffic separate from application requests.
+* *Use a firewall*: Use a firewall to shield production Open Liberty servers so that they don't run in the DMZ. The DMZ is an unsecured network segment that exists between internal and external networks. Firewalls help prevent security intrusions by monitoring network traffic and blocking specific traffic based on security rules.
+* *Enable the `transportSecurity` feature*: The feature:transportSecurity-1.0[] feature enables the feature:ssl-1.0[] feature to support SSL connections and adds the ability to configure outbound transports. In addition to enabling `transportSecurity-1.0`, disable non-SSL ports by setting them to `-1` in the `server.xml` file, and use only certificates that are signed by a certificate authority for public-facing production servers.
```
dmuelle: do we want to include version #s on features? Does the macro always grab the most current version? This should be consistent across the docs, whatever the call

ref/general/hardening-intro.adoc
```diff
@@ -22,49 +22,38 @@ Hardening is a general term that is used to describe the measures that are taken
...
+* *Run applications on a virtual host*: Configure applications to run on a virtual host. Open Liberty admin traffic comes in on the default host, so configuring your applications to run on a virtual host keeps Open Liberty admin traffic separate from application requests.
```
dmuelle: I know this is true for Liberty- does it apply to OL as well? Or does the webcontainer (eg Docker) take the place of the virtual host (or are they the same thing?) I'm not sure....

ref/general/hardening-intro.adoc
```diff
@@ -22,49 +22,38 @@ Hardening is a general term that is used to describe the measures that are taken
...
 Application configuration intrusions occur when external users run applications that derive or inherit privileges that they are not authorized to. One example of this type of intrusion happens when an application inherits the identity of the server, giving the application unauthorized permissions.
 
-To learn more about application configuration vulnerabilities and ways to prevent application intrusions, go to <link to application configuration topic>.
+To learn more about application configuration vulnerabilities and ways to prevent application intrusions, go to <link pending>.
 
-=== Best practices
+==== Best practices
 
-===== alskdjfa;
+* *Understand roles and priorities*: Roles and mappings are crucial for securely configuring your application. Incorrect role mapping can give inadvertent levels of access to users, which can result in some users with access to sensitive resources. You define roles by using annotations and the `web.xml` file, and you bind these roles to users or groups in the `ibm-application-bnd.xm` and `server.xml` files. For more information about roles and mapping them to the appropriate users or groups, go to <link pending>.
```
dmuelle: `...which can result in some users with access to sensitive resources.`

what about

`...which can result in users gaining unapproved access to sensitive resources.`

dmuelle: `You define roles by using annotations and the `web.xml` file, and you bind these roles to users or groups in the `ibm-application-bnd.xm` and `server.xml` files.`

This is just one way to specify and define roles- since you're not actually telling how to do this, might be better not to get into it here and just send them to the Authorization topic. I think you could omit this sentence (unless it was specifically requested) and the section would still work the same.

ref/general/hardening-intro.adoc
Charlotte-Holt: Updated

ref/general/hardening-intro.adoc
Charlotte-Holt: Okay, agreed.

ref/general/hardening-intro.adoc
Charlotte-Holt: Updated

Charlotte-Holt: Updated

ref/general/hardening-intro.adoc
Charlotte-Holt: Good point, and I'm not sure either. I'll ask Ajay when I send this first draft to him for review.

ref/general/hardening-intro.adoc
Charlotte-Holt: I'll _try_ to update it so that it resolves to the latest version, need to think about this in the context of Antora transition too

ref/general/hardening-intro.adoc
Charlotte-Holt: Updated.

Charlotte-Holt:
@dmuelle Thanks for the review! Incorporated the changes.

PR #1009, opened 2020-02-05
Update 11

ManasiGandhi:


ref/general/mp-openapi.adoc
```diff
@@ -10,47 +10,80 @@
...
+
+Developers can communicate with the REST API in their web service only if they have the relevant information to do so. It is important to document APIs when they are published so that other developers can use them.
+
+There are two main ways to generate a link:https://swagger.io/docs/specification/about/[OpenAPI] document. The developer can generate documentation of the REST API from the source code of the REST API itself. The reference document thus produced, lists all the API endpoints and descriptions of how to use them. Alternatively, the developer can start by designing the API in the form of structured documentation that they can then augment with annotations in the code. The developer can even generate the stubs for the source code, including client libraries for the API, from this manually created documentation.
```
rw2513: In the phrase, "The reference document thus produced", I think "thus" could be replaced with a more common word to ensure consistency and clarity. Maybe change to "The reference document is then produced".

ref/general/mp-openapi.adoc
```diff
@@ -10,47 +10,80 @@
...
+    @Operation(
+        summary = "List inventory contents.",
+        description = "Returns the currently stored host:properties pairs in the "
+        + "inventory.")
+    public InventoryList listContents() {
+        return manager.list();
+    }
+----
+
+In the following OpenAPI documentation example that is generated from the example code, the JAX-RS framework generates the path /inventory/systems based on the value of the @Path annotation, and the value of the operationId key (listContents) based on the name of the method (listContents()). The other parts of the document are generated from the OpenAPI annotations with which the developer annotated the code. For example, the @Operation annotation provides more meaningful information about what the operation does and why; in this case, the operation returns a list of the host names with properties of each host that are currently stored by the inventory web service. This is described by the summary and description keys in the @Operation annotation:
```
rw2513: In the phrase "the path /inventory/systems", put the noun after the path name, so change to "the /inventory/systems path". 
Change "OpenAPI annotations with which the developer annotated the code" to "OpenAPI annotations that the developer annotated the code with", this might help to make the wording a bit simpler

ref/general/mp-openapi.adoc
```diff
@@ -10,47 +10,80 @@
...
+              $ref: '#/components/schemas/InventoryList'
+----
+
+The information provided through the OpenAPI annotations augment the basic API documentation that is generated by the JAX-RS framework.
+
+For examples of annotations, see link:https://www.openliberty.io/docs/ref/microprofile/3.0/#package=org/eclipse/microprofile/openapi/annotations/package-frame.html&class=org/eclipse/microprofile/openapi/annotations/Components.html[Annotation Samples]
+
+== Manually created REST API documentation
+
+You can generate the static files before any code with an editor like Swagger editor. In this approach, popularly known as the 'design first' approach, the API contract is designed first before writing any code. The API contract acts as the central draft, often referred to as the "source of truth". This approach is considered efficient as it helps to spot issues in the design, before writing any code rather than after the implementation.
```
rw2513: Change "before writing any code" to "before you write any code". We typically want to avoid gerunds when possible to convey a more active voice.
There are two instances of this that I noticed in this section.

Issue #901, opened 2020-01-08
SSO configuration with JSON Web Tokens 


Charlotte-Holt:
From LC: This is a new concept topic that should:

- [ ] Explain all the Liberty JWT features and when to use them (e.g., someone doing SAML can ignore this topic) 
- [ ] Focus on what a developer needs to understand to use JWT - don't just document everything equally for the sake of being comprehensive. 
- [ ] Provide a scenario of securely propagating authorization with requests to other microservices (JWT) 

Info source: https://developer.ibm.com/wasdev/docs/using-signed-json-web-tokens-jwts-secure-microservices/
Proposed guide issue: https://github.com/OpenLiberty/guides-common/issues/131
Config examples w/ #636

Rwalls1:
@lauracowen In regards to the Liberty JWT features, is there a reference for what those are. I may be overlooking something but I can't seem to identify them.

lauracowen:
This is aimed at developers writing backend web services that need to be able to authorize requests to them (and requests from them to other web services) for information that is secured.

Start with the MicroProfile JWT capabilities (mpJwt feature) as that's the way we want/expect developers to go. Work through the JWT guide to learn about it: https://openliberty.io/guides/microprofile-jwt.html

The mpJwt feature depends on (automatically pulls in) the JWT feature, though you can also use the latter by itself (but start by focusing on developers using it through the MP JWT feature). There's a little bit about this (plus a full example you can take a look at) in this blog post: https://openliberty.io/blog/2019/08/29/securing-microservices-social-login-jwt.html

Talk to Bruce Tiffany about what is important for developers to know about in order to use JWT to secure their backend web services.

Rwalls1:
@brutif Please review: https://draft-openlibertyio.mybluemix.net/docs/ref/general/#/docs/concept/sso-config-json.html.

brutif:
@Rwalls1 looks good! Just a few minor suggestions- 

claims by it --> claims in it. 

trust the signature --> trust the issuer's signature

a secret key known to the OpenID Provider (OP) --> a secret key known to the creator of the JWT (anyone can create a JWT, not just OP's)

The net.oauth example used is ok, but that library is no longer maintained and has been removed from liberty.  You might say something like 

The following example shows the creation of a signed JWT with Java.  Multiple open source libraries for creating and consuming JWT's exist.  Liberty also provides API's to do this in the Jwt-1.0 feature. 

[...]

PR #1021, opened 2020-02-06
Authentication

dmuelle:
Edits per @lauracowen transcript of @NottyCode comments

- "user registries needs more than a para but not a full section heading just for federated user registry." 
**this section links out to separate topics for basic, LDAP and federated registries. Added info for custom registries. What else is missing from this section?**

- "JAAS needs to be included in the intro topic..."
**restored JAAS summary from previous draft**

- Diagram issues- current diagram doesnt make sense  
**what info/ scenarios should the diagram contain? - current sketch was adapted from JM security presentation slides & KC.. made a few changes but not sure if I'm getting any closer...**

- server vs app in intro
**rewrote opening sentence as passive per suggestion- does this work?**


ref/general/authentication.adoc
```diff
@@ -13,16 +13,20 @@
...
 
 For example, a user of a music streaming service might want to listen to a song, learn about the artist, and get recommendations for similar artists. Each of these tasks would require access to a different service, but having to authenticate for each of them separately uses unnecessary time and resources. Open Liberty supports a number of different strategies for single sign-on (SSO) authentication so that a user or service must authenticate only once to access the various resources in an application. For example, the link:/docs/ref/feature/#socialLogin-1.0.html[Social Media Login] feature enables users of an application that is running on Open Liberty to authenticate through their social media login. After authentication, an application can either obtain a JSON Web Token (JWT) from the social media service or associated OpenID Connect provider, or create a new one. This JWT can then be used to authenticate the user to any service in the application.
 
-image::/docs/img/Authentication.png[Authentication diagram,align="center"]
+image::/docs/img/Authentication-rvsd.png[Authentication diagram,align="center"]
```
NottyCode: I think we should not use upper case in file names.

ref/general/authentication.adoc
```diff
@@ -13,16 +13,20 @@
...
+A user registry is a database that contains information about a set of users and their associated roles. When a user requests a protected resource in an application, the user's credentials are checked against the information in one or more user registries. Open Liberty provides an easy-to-use link:/docs/ref/general/#basic-registry.html[basic user registry] for developers to use in test environments. Typically, applications in production are configured to use an external registry, such as a link:/docs/ref/general/#LDAP-registry.html[Lightweight Directory Access Protocol (LDAP) registry].
```
NottyCode: I'm not sure about the use of the term database in this context. The users are not necessarily stored in an SQL database which might be implied by the way this is expressed. I don't know of a better alternative.

ref/general/authentication.adoc
```diff
@@ -13,16 +13,20 @@
...
+
+Open Liberty also supports link:/docs/ref/general/#federated-registries.html[federated user registries] so that a server can be configured to authenticate from multiple registries from a single logical view. Information from LDAP, basic, and custom user registries can be federated into a common registry to manage application security.
```
NottyCode: I would probably swap this around and say that "Open Liberty supports multiple user registries and federates them into a single logical view." I think starting with what is possible and then introducing the term we use is probably easier to follow.

ref/general/authentication.adoc
```diff
@@ -13,16 +13,20 @@
...
+
+If your application needs to reference a database other than a basic or LDAP registry, you can configure a custom registry by implementing the `UserRegistry` interface. The `UserRegistry` interface is a Service Programming Interface (SPI) that enables support to virtually any type of account repository.
```
NottyCode: User Registry is an API. The S in SPI typically means System, not Service. What about using data store over database?

ref/general/authentication.adoc
```diff
@@ -34,18 +38,30 @@ Open Liberty supports the following options for SSO authentication:
 - link:/docs/ref/feature/#socialLogin-1.0.html[Social Media Login] +
 Users can authenticate to websites that are hosted on the Open Liberty server by logging in with their social media accounts. When the `socialLogin-1.0` feature is enabled, login with Facebook, Twitter, Google, Linkedin, and GitHub is automatically configured. You can also define your own configuration for any social media platform that is based on the OAuth 2.0 or OpenID Connect 1.0 standards.
 
+- link:/docs/ref/general/#/docs/concept/sso-config-json.html[JSON Web Tokens] +
+JSON Web Tokens (JWT) are used to propagate the user identity that is established by Single sign-on (SSO) between different microservice applications. By configuring SSO user authentication with JWT, you can manage the identity that is shared between applications.
```
NottyCode: I don't think this is right. It isn't describing what a JWT is, it is describing a usage pattern as if that is what it is. JWT is a way to represent an authenticated user identity, it can then be used between microservices. I don't think SSO is necessarily relevant.

ref/general/authentication.adoc
```diff
@@ -34,18 +38,30 @@ Open Liberty supports the following options for SSO authentication:
...
-OIDC is a popular internet SSO protocol that works well with cloud-native applications. This protocol enables a client application to request the identity of the user as an ID token in a standardized, REST-like manner. The client application, or Relying Party, requests information about a user from an authorization server, known as an OpenID Connect Provider. After users are authenticated by an OpenID Provider, they can access any Relying Party that is configured for that Provider, without having to manage unique accounts for each application or website.
+OIDC is a popular internet SSO protocol that works well with cloud-native applications. This protocol enables a client application to request the identity of the user as an ID token in a standardized, REST-like manner. The client application, or Relying Party, requests information about a user from an authorization server, which is known as an OpenID Connect Provider. After users are authenticated by an OpenID Provider, they can access any Relying Party that is configured for that Provider, without having to manage unique accounts for each application or website.
```
NottyCode: Is this first reference to OIDC? If so should the meaning be spelled out first? Not sure what changed here.

ref/general/authentication.adoc
```diff
@@ -34,18 +38,30 @@ Open Liberty supports the following options for SSO authentication:
...
 
 - link:/docs/ref/general/#/docs/concept/sso-config-saml.html[Security Assertion Markup Language (SAML 2.0)] +
 SAML Web Browser SSO enables web applications to delegate user authentication to a SAML identity provider instead of a configured user registry.
 
 - link:/docs/ref/feature/#spnego-1.0.html[Simple and Protected GSS-API Negotiation Mechanism (SPNEGO)] +
 With SPNEGO SSO, users log in to the Microsoft Domain controller once to access and move between protected resources on Open Liberty servers without getting prompted for additional logins.
 
+= JAAS authentication with Open Liberty
+
+Open Liberty relies on the Java Authentication and Authorization Service (JAAS) framework to determine a user's identity and access roles. JAAS authentication is suited to microservice-based applications because it separates the authentication process from the rest of the application so that authentication can be managed independently. New or updated authentication processes can be plugged in to an application without otherwise modifying the application itself.
```
NottyCode: This doesn't make sense to me.

ref/general/authentication.adoc
```diff
@@ -34,18 +38,30 @@ Open Liberty supports the following options for SSO authentication:
...
+
+The JAAS framework defines a `Subject` object to represent the source of a request to an application. The subject can be any entity that makes a request, such as a user or another microservice. The subject contains a set of principals, or details about the identity of the entity that made the request, which is used for authentication. The subject also contains sets of credentials, or security attributes that are used for authorization. By creating a subject to represent a user or service, Open Liberty provides an efficient and easily configurable means for authentication data to pass between microservices in an application.
```
NottyCode: I think this isn't right:

>The subject contains a set of principals, or details about the identity of the entity that made the request, which is used for authentication.
specifically I don't think the or should be there since you are describing what a principal is, or implies an alternative rather than a definition.

I think credentials are optional in a subject.

ref/general/authentication.adoc
```diff
@@ -34,18 +38,30 @@ Open Liberty supports the following options for SSO authentication:
...
+
+With JAAS in Open Liberty, each stage in the authentication process operates independently and can be updated or reconfigured with minimal impact on the other stages in the process. The `LoginContext` class determines the methods that are used for authentication, according to the `LoginModule` interfaces that are configured. Each `LoginModule` interface represents a particular type of authentication service. For example, one `LoginModule` interface can handle basic username and password authentication, while another processes SSO tokens.
```
NottyCode: This doesn't feel specific to Liberty, but the intro paragraph says it is.

ref/general/authentication.adoc
```diff
@@ -34,18 +38,30 @@ Open Liberty supports the following options for SSO authentication:
...
+OIDC is a popular internet SSO protocol that works well with cloud-native applications. This protocol enables a client application to request the identity of the user as an ID token in a standardized, REST-like manner. The client application, or Relying Party, requests information about a user from an authorization server, which is known as an OpenID Connect Provider. After users are authenticated by an OpenID Provider, they can access any Relying Party that is configured for that Provider, without having to manage unique accounts for each application or website.
 
 - link:/docs/ref/general/#/docs/concept/sso-config-saml.html[Security Assertion Markup Language (SAML 2.0)] +
 SAML Web Browser SSO enables web applications to delegate user authentication to a SAML identity provider instead of a configured user registry.
 
 - link:/docs/ref/feature/#spnego-1.0.html[Simple and Protected GSS-API Negotiation Mechanism (SPNEGO)] +
 With SPNEGO SSO, users log in to the Microsoft Domain controller once to access and move between protected resources on Open Liberty servers without getting prompted for additional logins.
 
+= JAAS authentication with Open Liberty
```
NottyCode: Not sure about the JAAS section, it feels overly verbose compared with the rest of the page.

ref/general/authentication.adoc
dmuelle: The UserRegistry/SPI info here comes from this KC page : https://www.ibm.com/support/knowledgecenter/SSAW57_liberty/com.ibm.websphere.wlp.nd.multiplatform.doc/ae/twlp_sec_custmr.html- 

which is basically implement SPI>convert to OSIG & bundle>install as usr feature.

If there's a User Registry API we should doc instead, or in addition to, is there autogen doc for it"? I can't find it but maybe I'm looking in the wrong places

For now, just replacing SPI w/ API

ref/general/authentication.adoc
dmuelle: should I remove this section altogether from the list of SSO options? It's based on this doc draft for the Jwt-1.0 feature : https://draft-openlibertyio.mybluemix.net/docs/ref/general/#/docs/concept/sso-config-json.html per #901

ref/general/authentication.adoc
dmuelle: not sure why there's a dif here but I added the acronym to the heading so it's defined for the description..

ref/general/authentication.adoc
dmuelle: there wasn't much conceptual info for JAAS in the KC resources so I went back to the Oracle spec to try and draft an overview of why JAAS should matter to  microservice developers using OL. I may have mis-conflated some aspects in the process. Also probably why this section seems out of joint with the rest. There was originally a ref to JAAS as "pluggable" here but a reviewer asked me to remove it. Combining/simplifying these two pps in this commit

ref/general/authentication.adoc
dmuelle: agreed- see comment above. Can this section get buy with less info? How much detail is relevant at this level of overview? Resources in KC go on for pages detailing LoginModules and CallbackHandlers etc. Here, just introducing the concepts, but is that still overkill in this context?

ref/general/authentication.adoc
NottyCode: I think replacing SPI with API in this instance is fine. UserRegistry has duel roles as both an SPI and an API, so in this case I think just going simple with API is fine.

The doc on writing OSGi bundle is out of date and no longer required for UserRegistry. It doesn't massively change the content in this document, but if we get to writing custom UserRegistry we should document the simpler approach.

ref/general/authentication.adoc
NottyCode: No, I think we should have a better description of what JWT is, rather than stating what it is used for. That should probably find its way into any description elsewhere of JWT.

dmuelle: @NottyCode new commit adds a description of what a JWT is..

ref/general/authentication.adoc
NottyCode: How about something like this:

>Open Liberty uses the Java Authentication and Authorization Service (JAAS) to authenticate users.
>JAAS makes use of named login contexts that are made up of a number of login modules that combine to authenticate a user.
>Open Liberty contains a number of built in login contexts and login modules which are used, but additional modules and contexts can be configured to customize authentication.
>During authentication the JAAS login modules authenticates the users credentials and loads user and group information.
>Depending on the kind of credentials provided the user and group information could be fetched from a UserRegistry, or an SSO token.

dmuelle: @NottyCode added this section with a few minor edits

PR #1023, opened 2020-02-06
edits per alasdair

dmuelle:
Edits per @lauracowen transcript of @NottyCode comments:

- "example of payroll application is not good" 
**reworked this example**

- "In some places, the wording is not inclusive enough of the alternative approaches. role names - some scenarios where a user registry is irrelevant (eg SSO). But if there’s a JWT, there’s no mapping to a user registry, the token just contains the groups." 
**brought some info in re tokenization/JWT as alternative to direct mapping from registry**

- " Don’t like the way it handles **. Dont’ think it should be simplified using the **…sentence. Should convey this info as it’s important but it’s not a “special value”; it’s a “special role name”. 
** edited this section and reordered it in the topic**

- "doc the simple scenarios first (group names and roles first - para 3; then para 4 - 
talk about roles, what roles are, then about the default mapping, then tell them how to do a non-default mapping"
**reorganized the roles sections**


ref/general/authorization.adoc
```diff
@@ -15,17 +15,18 @@
...
+In an application where only certain users or groups are entitled to access certain resources, users and groups are assigned roles that represent which resources they can access. For example, in a payroll application, users in the `Employee` role can view a page that lists their hourly wages and hours worked. However, only users in the `Manager` role can make changes to employee wages. During link:/docs/ref/general/#authentication.html[authentication], Open Liberty determines who a user is and what roles they are in. During authorization, Open Liberty determines how these roles correspond to different protected resources on the server.
+
+The roles that users and groups are checked against can either be defined directly in the application or applied to the application by way of the server configuration. It is also possible to avoid defining roles altogether if you specify role names that match the group names in the user registry exactly. In this case, the server  maps the role to the group with a matching name by default. In this configuration, you do not specify an `<application-bnd>` element in the `server.xml` file.
```
NottyCode: this is still not right. The roles are defined in the application, what can be specified in configuration is the mapping from a role to a user or group. The way this is written it makes it sound like the role itself is defined in server.xml. I would also not say to define the role names to match the groups, but the opposite way around. The role is the primary artifact.

ref/general/authorization.adoc
```diff
@@ -15,17 +15,18 @@
...
 Roles are defined directly in the application by setting the `@DeclareRoles` annotation or by using `<security-role>` elements in the `web.xml` deployment descriptor file.
 To apply roles to an application from the server configuration, specify the `<security-role>` element within the `<application-bnd>` element in the `server.xml` file.
 
-It is also possible to avoid defining roles altogether if you specify role names that match the group names in the user registry exactly. In this case, the server automatically maps the role to the group with a matching name. In this configuration, you do not specify an `<application-bnd>` element in the `server.xml` file. 
+In some use cases, an application needs to know only whether a user is authenticated or not. In such cases, configuration can be simplified by specifying `"**"` as the role name in the `@RolesAllowed` annotation or XML configuration. This value denotes that any authenticated user can access a resource. It is also possible to effectively bypass authorization altogether by using the `@PermitAll` annotation, which specifies that a resource is accessible to everyone, even users who are not authenticated.
```
NottyCode: PermitAll doesn't really bypass authorization, it is a kind of authorization that allows anyone. I know the distinction might seem nit picky, but I think it is better to think of PermitAll as an authorization rule that is very permissive rather than not having any.

ref/general/authorization.adoc
```diff
@@ -15,17 +15,18 @@
...
 
-In some use cases, an application needs to know only whether a user is authenticated or not. In such cases, configuration can be simplified by using the `"**"` special value in the `@RolesAllowed` annotation or XML configuration. This value denotes that any authenticated user can access a resource. It is also possible to effectively bypass authorization altogether by using the `@PermitAll` annotation, which specifies that a resource is accessible to everyone, even users who are not authenticated.
+Typically, a developer can specify which roles are allowed to access an application resource directly in the source code by setting the `@RolesAllowed` annotation to contain the permissible roles. When a user is authenticated, the access groups to which the user is assigned are retrieved from a user registry or single sign-on (SSO) provider and added to the `Subject` object for that user. The Open Liberty server maps these groups to the roles that are defined in the application. If the role or roles that are mapped to the user match the roles that are specified for the resource, the user is granted access.
```
NottyCode: This feels like it is describing internal implementation detail that really isn't relevant. I don't think we need to talk about a Subject her and the fact that we store user and group info in the Subject for role authorization is an implementation detail.

ref/general/authorization.adoc
dmuelle: rewrote this section top clarify, lmk if it works.

I changed the orientation between roles vs groups in this section to give roles precedence- but I'm confused. The default mapping occurs when roles match groups- but dont groups, esp as defined in an LDAP registry or by OIDC provider exist independent of any one application? Wouldn't an app dev have to know what user groups were defined in the existing security framework and then define roles to match?

NottyCode: There is a difference between how Role Based Access Control works from a design perspective, and how it is implemented from a performance perspective and I think the docs are written from an implementation perspective. 

A role is a collection of permissions, so a permission could allow you to read or write a file, or access a servlet, or an ejb. A role collects a bunch of disparate permissions together so you can manage them as a collection more easily. Then you "map" users or groups as having that role. The idea is some bit of code says "does the user have permission X" and at some point there is a mapping from a user and the groups the user is in to the roles those users and groups have and to the permissions. This can all be processing and memory intensive, so internally what tends to happen is the security check just says "does the user have this role" and we store the role data with the rest of the identity data. However this is an internal optimization and not really how we should describe RBAC.

From this perspective I think this is still worrying me:

> During link:/docs/ref/general/#authentication.html[authentication], Open Liberty determines who a user is and what roles they are in. During authorization, Open Liberty determines how these roles correspond to different protected resources on the server.

I think it probably should read like this:

> During link:/docs/ref/general/#authentication.html[authentication], Open Liberty identifies the user, what groups the user might be in, and which roles they have. When accessing a protected resource on the server this information is used to determine if the user is authorized to access the protected resource. 

For the second paragraph I think we need to simplify:

>The roles ~that users and groups are checked against~ are defined directly in the application. Out of the box the role names are mapped directly to group names in the user registry or single sign-on token. Often group names do not directly map to role names as defined in the application. In this case a mapping from role names to users and groups can be defined in server.xml.

dmuelle: @NottyCode pushed a new commit to make these changes

PR #1076, opened 2020-02-13
Update Security vulnerabilities list

clarkek123:


clarkek123:
This change removes the comment from the Notes field for CVE-2019-4720 to remove the comment about affecting transportSecurity-1.0 since this is part of liberty core. This column will be blank.

PR #1084, opened 2020-02-13
Recreated example and description files for OIDC server

ManasiGandhi:


ref/feature/openidConnectServer/description.adoc
```diff
@@ -0,0 +1,5 @@
+Single sign-on is a process to sign in once with your credentials to access multiple applications.
```
rw2513: For consistency purposes, I think you should reference the abbreviation for Single sign-on first here so change "Single sign-on to Single sign-on (SSO).

ref/feature/openidConnectServer/examples.adoc
```diff
@@ -0,0 +1,32 @@
+The OpenID Connect Provider (openidConnectServer-1.0) feature can be configured to manage user authentication for OpenID Connect clients and microservices during development.
+The OpenID Connect Provider can work directly with a user registry such as LDAP or act as an identity broker and delegate to another security provider.
```
rw2513: If the OP abbreviation that you in the phrase, "Add the OP Open Liberty feature" stands for OpenID Connect Provder, then I think you should change "The OpenID Connect Provider to "The OpenID Connect Provider (OP)" to avoid any possible confusion

PR #1129, opened 2020-02-24
OIDC server config snippet updates

ManasiGandhi:


ref/feature/openidConnectServer-1.0.adoc
```diff
@@ -1,12 +1,14 @@
...
+:linkcss:
 :page-layout: feature
-:nofooter: 
+:nofooter:
 
 include::./gen/openidConnectServer-1.0.gendoc[tag=description]
 
 include::./gen/openidConnectServer-1.0.gendoc[tag=enable]
 
+include::./gen/openidConnectServer-1.0./examples.adoc[]
```
Charlotte-Holt: I think this should be:
`include::./openidConnectServer/examples.adoc[]`

(Never done this before, but going off the example from https://raw.githubusercontent.com/OpenLiberty/docs/master/ref/feature/jdbc-4.0.adoc.)

Also, do you need to add the description file too?

ManasiGandhi: Yes. Need to add the description file but need to discuss before I do so as not sure.

ref/feature/openidConnectServer-1.0.adoc
```diff
@@ -1,12 +1,17 @@
...
+:linkcss:
 :page-layout: feature
-:nofooter: 
+:nofooter:
 
 include::./gen/openidConnectServer-1.0.gendoc[tag=description]
 
 include::./gen/openidConnectServer-1.0.gendoc[tag=enable]
 
+<<<<<<< Updated upstream
```
lauracowen: Don't think these git diff lines should still be in the file.

ref/feature/openidConnectServer-1.0.adoc
```diff
@@ -1,12 +1,17 @@
...
+include::./openidConnectServer/examples.adoc[]
+
+=======
```
lauracowen: See above but think this git diff line should be removed.

ref/feature/openidConnectServer-1.0.adoc
```diff
@@ -1,12 +1,17 @@
...
+>>>>>>> Stashed changes
```
lauracowen: See above but think this git diff line should be removed.

ref/feature/openidConnectServer-1.0.adoc
ManasiGandhi: Removed.

ref/feature/openidConnectServer-1.0.adoc
ManasiGandhi: Removed.

ref/feature/openidConnectServer-1.0.adoc
ManasiGandhi: Removed.

ref/feature/openidConnectServer-1.0.adoc
ManasiGandhi: Removed

PR #1139, opened 2020-02-25
Open api edits

ManasiGandhi:


ref/general/mp-openapi.adoc
```diff
@@ -12,19 +12,29 @@
...
+The reference document then produced, lists all the API endpoints and descriptions of how to use them.
+Alternatively, the developer can start by designing the API in the form of structured documentation that they can then augment with annotations in the code.
+The developer can even generate the stubs for the source code, including client libraries for the API, from this manually created documentation.
 
 == Processing annotations
 
 You can augment the existing code with OpenAPI annotations, which provide detailed documentation for the REST API.
 
-For example, the following code example shows JAX-RS code to request the list of host names in the inventory. The code has been annotated with OpenAPI annotations to provide more useful and explanatory information about the operation:
+For example, the following code example shows JAX-RS code to request the list of host names in the inventory.
```
rw2513: I dont think you need to reference "code" or "example" twice here so I think you can simplify this sentence by either changing this sentence to something like "The following example shows a JAX-RS request to list host names in the inventory." I just don't think you need to have both "For example" and the "following code example" here to reference the code snippet

ref/general/mp-openapi.adoc
```diff
@@ -12,19 +12,29 @@
...
+The code has been annotated with OpenAPI annotations to provide more useful and explanatory information about the operation:
```
rw2513: The phrase "The code has been annotated" might sound a bit too passive. I think maybe change the entire sentence to "The OpenAPI annotations in the example provide more useful and explanatory information about the operation:"

ref/general/mp-openapi.adoc
```diff
@@ -47,7 +57,10 @@ For example, the following code example shows JAX-RS code to request the list of
     }
 ----
 
-In the following OpenAPI documentation example that is generated from the example code, the JAX-RS framework generates the path /inventory/systems path based on the value of the @Path annotation, and the value of the operationId key (listContents) based on the name of the method (listContents()). The other parts of the document are generated from the OpenAPI annotations that the developer annotated the code with. For example, the @Operation annotation provides more meaningful information about what the operation does and why; in this case, the operation returns a list of the host names with properties of each host that are currently stored by the inventory web service. This is described by the summary and description keys in the @Operation annotation:
+In the following OpenAPI documentation example that is generated from the example code, the JAX-RS framework generates the `path /inventory/systems` path based on the value of the `@Path` annotation, and the value of the `operationId` key (listContents) based on the name of the method (listContents()).
```
rw2513: With the phrase "that is generated from the example code", I think maybe replace "the example code" with a reference to what the example code is so if the example code is a JAX-RS request then change it to something like "that is generated from the JAX-RX request". This can help highlight exactly what the user needs to know about the example.

ref/general/mp-openapi.adoc
```diff
@@ -47,7 +57,10 @@ For example, the following code example shows JAX-RS code to request the list of
...
+The other parts of the document are generated from the OpenAPI annotations that the developer annotated the code with.
+For example, the `@Operation` annotation provides more meaningful information about what the operation does and why; in this case, the operation returns a list of the host names with properties of each host that are currently stored by the inventory web service.
+This is described by the summary and description keys in the `@Operation` annotation:
```
rw2513: Change "This" to reference what is being described to avoid ambiguity. So maybe change it something like "This action" or whatever is being described by the summary and description keys.

ref/general/mp-openapi.adoc
```diff
@@ -12,19 +12,27 @@
...
+The structured documentation helps other developers and microservices understand and communicate with your microservice.
+Developers can write a web service to communicate with the REST API of another web service only if they have the relevant information to do so.
+MP OpenAPI specification facilitates this communication by generating human and machine readable documentation on an easy to use interface without the need to access the source code.
 
-Developers can communicate with the REST API in their web service only if they have the relevant information to do so. It is important to document APIs when they are published so that other developers can use them.
+There are two main approaches to generate an link:https://swagger.io/docs/specification/about/[OpenAPI] document, code-first and design-first.
+In the code-first approach, the developer can generate documentation of the REST API from the source code.
+The reference document then produced, lists all the API endpoints and descriptions of how to use them.
```
Rwalls1: This sentence reads a bit awkwardly. Maybe change "The reference document then produced lists all" to "A reference document is then produced that lists all". Also, for "descriptions of how to use them", I think it may be better to clarify what "them" is referring to in order to avoid any possible confusion.

PR #1181, opened 2020-03-04
add examples to fed reg feature doc

dmuelle:


ref/feature/federatedRegistry/examples .adoc
```diff
@@ -0,0 +1,60 @@
+
+== Examples
+
+=== Federation of basic registries
+
+To federate two or more basic registries, you must enable both link:/docs/ref/feature/#appSecurity-3.0.html[the Application Security feature] and the Federated User Registry feature, as shown in the following example.
```
Charlotte-Holt: You may want to put a colon after "following example" rather than a period.

dmuelle:
Thanks, fixed

PR #1206, opened 2020-03-10
Origin/request timing

ManasiGandhi:


ref/general/requestTiming.adoc
```diff
@@ -0,0 +1,71 @@
+:page-layout: general-reference
+:page-type: general
+:page-description: The request timing feature automatically detects slow requests and logs the information when the request is completed.
+:page-categories: MicroProfile
+:seo-title: The request timing feature automatically detects slow and hung requests
+:seo-description: The request timing feature automatically detects slow and hung requests and logs the information when the request is completed.
+= Slow and hung request detection
+
+Slow or hung requests can happen, and your application needs a continuous monitoring tool to detect the cause.
```
dmuelle: the first sentence seems fragmented, and doesn't tell you what a slow or hung request is.

"When an application takes longer than expected to complete a request, the request can be logged as slow or hung. Your applications need a continuous monitoring tool to detect and analyze slow and hung requests."

or something like that....

ref/general/requestTiming.adoc
```diff
@@ -0,0 +1,71 @@
...
+The request timing feature automatically detects slow  and hung requests and logs the information when the request is completed.
```
dmuelle: should Request Timing be capitalized? I think feature names are capitalized...

ref/general/requestTiming.adoc
```diff
@@ -0,0 +1,71 @@
...
+
+The `requestTiming` feature monitors requests with respect to time and automatically detects and diagnoses slow and hung requests.
```
dmuelle: use plain text name. If you have to use the monospace/code name, include the version number- but avoid that if possible

ref/general/requestTiming.adoc
```diff
@@ -0,0 +1,71 @@
...
+In addition to the previous capability of detecting the slow request, the updated feature now also logs when the request is completed, thus improving the monitoring efficiency.
```
dmuelle: I don't think we need to talk about the previous version unless the doc is announcing the new version. In this case, just state that the feature detects the slow request and logs completion

ref/general/requestTiming.adoc
```diff
@@ -0,0 +1,71 @@
...
+The users can view the information about a completed slow request with the request id and time of completion.
```
dmuelle: maybe don't switch to "users" here- as there is some ambiguity about end user of an app vs developer. Stay with "You can view..."

Does this sentence mean that you need to know the request id and time of completion to view info about a request? It's not clear

ref/general/requestTiming.adoc
```diff
@@ -0,0 +1,71 @@
...
+Similarly, in the case of a hung request, a message is logged if a request hung initially, gets completed.
```
dmuelle: I think this means-

"If a message that was flagged as hung later completes, the completion is recorded in the logs"

but im not sure

ref/general/requestTiming.adoc
```diff
@@ -0,0 +1,71 @@
...
+
+== Configure slow and hung request detection
+
+You can enable the request timing the feature by adding the `requestTiming` to the list of features in the `server.xml` file.
```
dmuelle: include feature version with monospace name

ref/general/requestTiming.adoc
```diff
@@ -0,0 +1,71 @@
...
+[source,java]
+----
+<featureManager>
+      ...
+      <feature>requestTiming-1.0</feature>
+      ...
+</featureManager>
+----
+
+The slow and hung request thresholds are configured by setting `slowRequestThreshold` and `hungRequestThreshold` in the `<requestTiming/>` configuration in the `server.xml` file.
```
dmuelle: The slow and hung request thresholds are configured by setting the `slowRequestThreshold` and `hungRequestThreshold` attributes on the `requestTiming` element in your `server.xml` file.

ref/general/requestTiming.adoc
```diff
@@ -0,0 +1,71 @@
...
+This setting sets a single slow and hung request threshold for all requests that are timed.
```
dmuelle: not sure single is correct because each attribute sets a threshold- so there should be two?

These attributes specify global slow and hung request thresholds for all requests that are timed.

ref/general/requestTiming.adoc
```diff
@@ -0,0 +1,71 @@
...
+
+== Detect slow requests
+
+When a request runs for longer than the `slowRequestThreshold` or any configured threshold, the `requestTiming` feature generates a warning message in the messages log file.
```
dmuelle: use plain text feature name

ref/general/requestTiming.adoc
```diff
@@ -0,0 +1,71 @@
...
+The message includes details about the request and events that made up the request.
+The default threshold value is 10 seconds.
+
+Following is an example of the format of the log messages.
```
dmuelle: Don't use "following" as the subject

The following example shows the format of a log message:

ref/general/requestTiming.adoc
```diff
@@ -0,0 +1,71 @@
...
+
+[source,java]
+----
+TRAS0112W: Request <(Request ID)> has been running on thread <THREADID> for at least <DURATION>. The following stack trace shows what this thread is currently running.
+<STACK TRACE>
+<DURATION AND OPERATIONS Table>
+----
+
+The message helps you analyze the reason behind the slow request, though it is difficult to determine whether the request is stuck at that point or is still running slowly.
```
dmuelle: should it say "hung" instead of "stuck"?

ref/general/requestTiming.adoc
```diff
@@ -0,0 +1,71 @@
...
+To address this issue, three messages are displayed for any slow request at the interval of specified `<slowRequestThreshold>`.
+The three different stack trace and request data gives you a better insight into the issue.
```
dmuelle: what is stack trace data? is it part of the message content? Might want to explain what is in the messages in a sentence or two. If you do use the word "data", remember that it is plural so should be  

The three different stack trace and request data _give_ you 

But i think you need to say why or how insight is improved

ref/general/requestTiming.adoc
```diff
@@ -0,0 +1,71 @@
...
+
+No further warnings are logged about the request unless the duration of the request crosses the hung request detection threshold.
+
+== Detect hung requests
+
+If the request exceeds the default `hungRequestThreshold` or the configured threshold, a warning message is generated in the messages log file with the details about the request and events that made up the request.
+A series of three thread dumps (javacores) is taken with a 1-minute delay between them.
```
dmuelle: should probably avoid any insinuation of "taking dumps" :)

maybe "is issued" ?

ref/general/requestTiming.adoc
```diff
@@ -0,0 +1,71 @@
...
+Further thread dumps are created only if the new requests are detected to be hanging.
```
dmuelle: Further thread dumps are created only if new hung requests are detected.

ref/general/requestTiming.adoc
ManasiGandhi: The Request Timing feature will automatically detect and provide the information about the request id and time of completion. Updated the doc.

ref/general/requestTiming.adoc
ManasiGandhi: How about 'Also, if a previously hung request is completed, a detailed message is logged, with the ID and time of completion of the request.'

dmuelle: Sounds good

ref/general/requestTiming.adoc
```diff
@@ -0,0 +1,71 @@
...
+
+[source,java]
+----
+TRAS0112W: Request <(Request ID)> has been running on thread <THREADID> for at least <DURATION>. The following stack trace shows what this thread is currently running.
+<STACK TRACE>
+<DURATION AND OPERATIONS Table>
+----
+
+The message helps you analyze the reason behind the slow request, though it is difficult to determine whether the request is hung at that point or is still running slowly.
+To address this issue, three messages are displayed for any slow request at the interval of specified `<slowRequestThreshold>`.
```
dmuelle: change
at the interval of specified `<slowRequestThreshold>`.
to
at the interval specified by the `<slowRequestThreshold>` attribute.

PR #1211, opened 2020-03-11
Update microshed.adoc

ManasiGandhi:


ref/general/microshed.adoc
```diff
@@ -6,78 +6,87 @@
...
+Developers mostly use different tools in development as compared to production for speed and ease of use.
+For example you might use a local maven build in development, but the application might be deployed to a docker container in production.
 Differences between the environments can cause a test to fail in production, even though it passed in development.
-You can address this issue by testing your application in the same container that you use in production.
-
-Running tests in development mode is convenient for local development, but it can be tedious to test against a running Open Liberty server in production scenarios, such as CI/CD pipelines.
-For this reason, MicroShed Testing can start and stop the application runtime before and after the tests are run.
-
-== How it works
+The application might work on your machine but not in production.
```
dmuelle: However, the only way to ensure that what works in your local environment also works after deployment is to use the same services in development as you do in production.

ref/general/microshed.adoc
```diff
@@ -6,78 +6,87 @@
...
-
-<!-- Any compatible version of JUnit Jupiter 5.X will work -->
-<dependency>
-    <groupId>org.junit.jupiter</groupId>
-    <artifactId>junit-jupiter</artifactId>
-    <version>5.4.2</version>
-    <scope>test</scope>
-</dependency>
-----
+The programming paradigms have changed with how we test software.
```
dmuelle: New programming paradigms are changing how we test software.

is that what you mean? As written, it sounds like the tests changed the paradigm

ref/general/microshed.adoc
```diff
@@ -6,78 +6,87 @@
...
+Currently, many developers run containers in production.
+Containers are convenient and combine everything you need to run a service into a single unit.
+They are easy to manage with an orchestration software like Kubernetes.
+They are lightweight and disposable and fairly easy to configure with environment variables.
```
dmuelle: They are lightweight, disposable, and fairly easy to configure with environment variables.

is "disposable" correct? I'm not familiar with it in this context. Or is it "composable"?

ref/general/microshed.adoc
```diff
@@ -6,78 +6,87 @@
...
+They are easy to leverage containers in a test environment using testcontainers, which is a Java library that supports Junit tests.
```
dmuelle: They are easy to leverage containers in a test environment

what does "they" refer to? Looks like "Conatiners are easy to leverage containers in a test environment..."

ref/general/microshed.adoc
```diff
@@ -6,78 +6,87 @@
...
+Testcontainers wrap the docker API.
+Microshed testing is a Java library for Microprofile and Java EE developers to test their applications similar to the production environment.
```
dmuelle: ...to test their applications similar to the production environment.

to test their applications in an environment similar to production.

ref/general/microshed.adoc
```diff
@@ -6,78 +6,87 @@
...
 
-Next, create a test class with the `MicroShedTest` annotation.
-Then, create a public `static MicroProfileApplication` method and inject one or more `public static JAX-RS` resources, as shown in the following example:
+Using MicroShed Testing, we can write an integration test that looks something like this:
```
dmuelle: should this be "you" instead of "we" ?

ref/general/microshed.adoc
```diff
@@ -6,78 +6,87 @@
...
+Microshed testing makes your work easier and faster by streamlining integration tests with testcontainers.
 
-Integration testing looks at microservices as a group and tests them as a whole unit.
-Integration testing offers advantages over other types of testing, for example, detecting bugs earlier in the cycle.
 
 == Why integration testing
 
-Integration testing follows unit testing.
-After individual modules are unit tested, integration testing looks at the modules together to evaluate their performance as a group.
+You need to be sure that your application works in development as well as production mode.
```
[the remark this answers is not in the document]
ManasiGandhi: I added this sentence based on a presentation slide by Andrew which uses the word 'confident'. I used 'sure'. From a developer point of view a test can run in development but fail in production, so the developers needs to be sure/confident about their test in developement.

ref/general/microshed.adoc
ManasiGandhi: It kind of means that paradigms have changed about how we test software because there it is no more needed to use different test tools in  development and production. But I took that sentence away.

ref/general/microshed.adoc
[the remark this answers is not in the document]
dmuelle: ok- but I think the sentence construction is implying am emphasis on development rather than production, while this section emphasizes vulnerabilities to production. eg

You need to be sure that your applications works as well in production as it does in the development environment.

ref/general/microshed.adoc
dmuelle: disposable is correct

ref/general/microshed.adoc
```diff
@@ -6,78 +6,87 @@
...
+Therefore, developers find integration testing to be more reliable for their application to work in production environment.
+However, integration tests take longer and can be complicated to set up and configure.
+With faster development cycles and limited time available to developers, it's not always feasible to run integrated tests for your applications.
 
-== What challenges does MicroShed resolve
+== Development-production parity
 
-An application might run differently in production as compared to the development and testing environment.
+Development-production parity, or dev-prod parity, is one of the factors in the link:https://12factor.net/[twelve-factor app], a methodology to build modern applications.
+The idea behind dev-prod parity is to keep the development, staging, and production environments as similar as possible, and it addresses gaps in development ad production in terms of time, personnel, and tools.
```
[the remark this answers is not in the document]
dmuelle: another possibility-

The idea behind dev-prod parity is to keep the development, staging, and production environments as similar as possible  in terms of time, personnel, and tools.

PR #1278, opened 2020-03-25
Add examples file for SPNEGO feature doc

Charlotte-Holt:


ref/feature/spnego/examples.adoc
```diff
@@ -0,0 +1,27 @@
+== Examples
+
+Open Liberty supports many common SPNEGO scenarios and configurations.
+
+=== Limit SPNEGO authentication 
+You can filter HTTP requests to require SPNEGO authentication for only certain requests, web applications, hosts, or user agents.
+
+The following example configures the following things:
```
ManasiGandhi: 'The following example configures the following things:'
Maybe rephrase this sentence. How about 'the example provided configures the following things/points'?

ref/feature/spnego/examples.adoc
```diff
@@ -0,0 +1,27 @@
...
+<1> SPNEGO authentication is used for any requests that are received containing the `example.com` hostname for resources within the `protectedApp` web application.
```
ManasiGandhi: "SPNEGO authentication is used for any requests that are received containing the `example.com` hostname for resources within the `protectedApp` web application."
to 
"SPNEGO authentication is used for any requests received that contain the `example.com` hostname for resources within the `protectedApp` web application."

ref/feature/spnego/examples.adoc
```diff
@@ -0,0 +1,27 @@
...
+<2> The client's GSS credentials are not added to the user subject upon successful authentication.
+<3> The Kerberos configuration and keytab files to be used by the server are given specific locations within the server configuration directory instead of their respective default locations.
```
ManasiGandhi: 'keytab files' should there be backticks for `keytab`?

Charlotte-Holt: From what I was able to find, I don't think it does.

ref/feature/spnego/examples.adoc
Charlotte-Holt: Updated

ref/feature/spnego/examples.adoc
Charlotte-Holt: Yeah, I considered this too, but I don't think that I should use a slash between two words, so I've left as is.

PR #1283, opened 2020-03-25
Tls config examples

dmuelle:


dmuelle:
I believe this one and #1264 (Federated reg config) are the only two ready to go atm. There should be a few more in the coming days- stuff that will go into Red Hat as well

PR #1384, opened 2020-04-09
samlWeb draft

ManasiGandhi:


ref/feature/samlWeb/examples.adoc
```diff
@@ -0,0 +1,33 @@
...
+
+•	AssertionConsumerService URL:
+
+`https://<hostname>:<sslport> /ibm/saml20/defaultSP/acs`
+
+•	Service Provider (SP) metadata URL:
+
+`https://<hostname>:<sslport> /ibm/saml20/defaultSP/samlmetadata`
+
+This URL is used to download the metadata for the service provider (SP) and provide the URL to the SAML identity provider to establish federation between this SP and Identity Provider (IdP).
```
dmuelle, line 21: `This URL is used to download the metadata for the service provider (SP) and provide the URL to the SAML identity provider to establish federation between this SP and Identity Provider (IdP).`

- Is `This URL..` the same URL as `provide the URL to the SAML identity provider...` It's not clear
- is `the SAML identity provider` the same entity as `Identity Provider (IdP)` ? Is so, abbreviate on first mention and use the abbreviation on the second mention. If not, needs to be more clear what the difference is.
- can this sentence be in active voice? ie state _what_ is using the URL to dl metadata and make that the subject of the sentence

ref/feature/samlWeb/examples.adoc
```diff
@@ -0,0 +1,33 @@
...
+
+The `samlWebSso20` element is configured by default in Liberty.
+
+----
+<samlWebSso20 id="defaultSP">
+
+</samlWebSso20>
+----
+
+The following default values are assumed in this default configuration,
```
dmuelle, line 11: suggested edit:

The default configuration sets the following values:

ref/feature/samlWeb/examples.adoc
```diff
@@ -0,0 +1,33 @@
...
+`https://<hostname>:<sslport> /ibm/saml20/defaultSP/acs`
+
+•	Service Provider (SP) metadata URL:
+
+`https://<hostname>:<sslport> /ibm/saml20/defaultSP/samlmetadata`
+
+This URL is used to download the metadata for the service provider (SP) and provide the URL to the SAML identity provider to establish federation between this SP and Identity Provider (IdP).
+
+The IdP metadata file must be copied to the `resources/security` directory on the server, and named `idpMetadata.xml`.
+The issuer name for SAML assertion is used as the security realm, and NameID is used as the principal to create an authenticated subject from the SAML assertion.
```
dmuelle, line 24: What is `NameID` ? An attribute? parameter? Add a descriptor and make monospace

ref/feature/samlWeb/examples.adoc
```diff
@@ -0,0 +1,33 @@
...
+
+The SAML `AuthnRequest` is signed with a private key in the default keystore of the server if the attribute `KeyStoreRef` is not specified. If the `keyAlias` is not configured, then samlsp is the default key alias.
```
dmuelle, line 26: - what is `AuthnRequest`- an attribute? 
- change "the attribute `KeyStoreRef`" to "the `KeyStoreRef` attribute "
- Is `keyAlias` an attribute?
- should `samlsp` be in monspace? Is it a value of the key alias attribute? ie "then `samlsp` is the default value of the key alias" If that's a correct interpretation...

ref/feature/samlWeb/examples.adoc
```diff
@@ -0,0 +1,33 @@
...
+If the `keyAlias` is not configured, and the keystore only contains one private key, the private key is used in the signature.
```
dmuelle, line 27: - If `keyAlias` attribute...
- change "the keystore only contains one private key" to "the keystore  contains only one private key"

ref/feature/samlWeb/examples.adoc
```diff
@@ -0,0 +1,33 @@
...
+
+If you create a new service provider instance, and the defaultSP is no longer required, then the defaultSP instance must  be explicitly disabled, by adding the following to the `server.xml` file:
```
dmuelle, line 29: - what is defaultSP?
- **change** If you create a new service provider instance, and the defaultSP is no longer required, then the defaultSP instance must  be explicitly disabled, by adding the following to the `server.xml` file:
**to**
If you create a new service provider instance and the `defaultSP` instance is no longer required, then you must explicitly disable the `defaultSP` instance. You can disable the defaultSP instance by adding the following code to the `server.xml` file:

ref/feature/samlWeb/examples.adoc
ManasiGandhi: I think SAML is the service provider here.

PR #1408, opened 2020-04-14
Add MP Metrics 2.3 info

Charlotte-Holt:


ref/general/metrics-catalog.adoc
```diff
@@ -413,7 +413,8 @@ In the table, the Prometheus metric names are listed after each metric. The tabl
  {empty}+
  base_REST_request_elapsedTime_seconds{class="<fully_qualified_class_name>",method="<method_signature>"} / (seconds)
 |Simple Timer
-|The number of invocations and total response time of the RESTful resource method since the start of the server.
+|This metric is available in link:https://openliberty.io/docs/ref/feature/#mpMetrics-2.3.html[MicroProfile Metrics 2.3] and later.
```
dmuelle, line 416: most of the other fields have a definition of the metric, followed by an availability statement. The order is flipped here- isn that intentional? Also- is there anything else about what context this metric is available in that the user would need to know? The other metrics state availability in the context of annotations used, rather than feature version. Good to highlight feature version here, but is that all?

PR #1409, opened 2020-04-14
Add MP Metrics 2.3 info to concept topic

Charlotte-Holt:


ref/general/microservice-observability-metrics.adoc
```diff
@@ -83,6 +83,7 @@ public String donateAmountViaCreditCard(@FormParam("amount") Long amount, @FormP
 '''
 
 === Simple Timers
+This type of metric is available link:https://openliberty.io/blog/2020/04/09/microprofile-3-3-open-liberty-20004.html#mra[beginning in MicroProfile Metrics 2.3].
 A simple timer metric tracks the elapsed timing duration and invocation counts.
```
dmuelle, line 87: Should light-weight be lightweight? Usually one word. ? Wondering here as well if definition should go first, possibly on it's own line

PR #1419, opened 2020-04-15
cleaning up

dmuelle:


dmuelle:
# Peer review

## Short desc

`The securityUtility command supports plain text obfuscation and TLS certificate creation for Open Liberty. You can call the tool directly from the command line.`

Inn the first sentence securityUtility is a command. In the second, it's referred to as a tool. Should there be just one term?

## Usage examples

`Create a TLS certificate for the myserver server that is valid for 365 days with the mypassword password:`

The example also includes the `subject` option. Is this option required to create a certificate? If so, it needs an explanation. If not, needs an explanation and mention that it's optional. I see that it is defined later in the topic so a full explanation isn't necessary here. But mention it so it's not the only option not mentioned in the description. You might also preface the user examples with some reference to option explanations later in the topic.

`Get help information about the createSSLCertificate action:`

Does the `help` option work for any action, or just `createSSLCertificate` ? If it works for any action, you might explain what the option does in a sentence, then intro this an example in the following sentence.

Should the description of interactive mode from the Encode options section be moved here? or duplicated? Otherwise you reference interactive mode without having explained it.

## Syntax

should this section appear before the usage examples? It provides info that helps in understanding the examples. Also, as is it is seems like another example, rather than a syntax template

This section also mixes "command" and "tool" to mean the same thing. Do they mean the same thing? If so, use one term.

`One or more optional parameters that are different based on the value of action.`

Should this say what options do, rather than just that they are different? For example

`One or more optional parameters that you can specify to modify an action.`

## Actions and options
`Use the following actions with the securityUtility command to get help information, encode passwords, and generate TLS certificates. Some actions can also be used with optional parameters, depending on what you need to accomplish.`

should it be "obfuscate" rather than encode?
Do you need `depending on what you need to accomplish.`? I think it's implied by "optional"

`This action prints help information for a specified action.`

Is "print" the preferred term in this context? I know I've seen it used w/r/t displaying text in the CLI, but there is also a specific CLI `print` command- https://en.wikipedia.org/wiki/PRINT_(command) "displays" might be less ambiguous. I see "display" used elsewhere in this topic .

## encode
`Interactive mode prompts you to enter and confirm your text input, rather than providing text on the command line. Otherwise, the provided text is encoded.`

"rather than providing text on the command line" should this be rather than providing text along with the command" or "on the same line as the command"? Since when the user is prompted, they still technically have to enter the text on the command line (unless I'm misunderstanding the function...)

Should it be "the provided text is obfuscated." ?

 `Supported encodings are..`
should it be  `Supported encoding types are..` for continuity with the preceding example....

 `The key can be provided to the server by defining the wlp.password.encryption.key variable whose value is the key.`

Does this mean the user needs to specify the key as the value of the varibale? Or does the variable retrieve the key value? As written, it sounds like the latter. If that's the case, should still probably revise to remove "whose" eg. "the value of which is the key". 

If the user has to explicitly specify the value, could be 
`You can provide the server with the key  by defining the wlp.password.encryption.key variable with the key as the value.`  or similar

`It is optional to provide password text on the command line` -see above comment re interactive mode

## createSSLCertificate
`Specifies the name of the Open Liberty client for which the keystore and certificate is created.`
should be "are created"

`The PKCS12 format is frequently the format that certificate authorities provide when they issue certificates.` 
suggested revision:
`Certificate authorities frequently issue certificates in the PKCS12 format.`

`The key can be provided to the server by defining the wlp.password.encryption.key variable whose value is the key. ` see previous comment re "whose value"

`If you use certificates that are created with EC keys, then include a customized ciphers list in your TLS configuration to include EC ciphers.`
should this be
`If you use certificates that are created with EC keys, then include a customized ciphers list in your TLS configuration that includes EC ciphers.`

`If the --extInfo option is used, the value is passed directly to the keytool command. Use the syntax of the --ext flag on the keytool command.`

What is the `keytool` command? How would someone get info on its syntax?

PR #1437, opened 2020-04-17
Updates to packaging and deploying topic

Charlotte-Holt:


ref/general/packaging-deploying.adoc
```diff
@@ -14,79 +14,69 @@
 = Package and deploy applications with Open Liberty
 
 You can choose among different options for packaging, containerizing, and deploying applications with Open Liberty.
-However, the simplest workflow for developing your cloud-native applications is to build applications with Maven or Gradle, containerize server packages with Docker, and deploy to the cloud with Kubernetes or OpenShift.
+However, the simplest workflow for developing your cloud-native applications is to build applications with Maven or Gradle, containerize with Docker, and deploy to the cloud with Kubernetes or OpenShift.
```
NottyCode, line 17: I am not keen on starting this sentence with however. It feels kind of negative. Also the prior sentence references containerizing, but the options being sort of excluded here include not containerizing.

ref/general/packaging-deploying.adoc
```diff
@@ -14,79 +14,69 @@
...
-Before you containerize and deploy your application, package it with the Open Liberty runtime.
-There are different tools that you can use to package your application.
-Using Maven or Gradle to build and package your application is one of the easiest ways to get your application ready for cloud deployment.
-Maven and Gradle both work well as part of continuous integration and continuous deployment (CI/CD) pipelines.
-They start with the Open Liberty kernel and download specific feature versions that are listed as dependencies.
-
-=== Maven
+== Maven and Gradle build tools
 link:https://maven.apache.org/[Maven] is a build automation tool that makes managing your builds and dependencies easier.
 With Maven, you use a Project Object Model (POM) file written in XML to define information about your project, and Maven downloads and installs all dependencies for you.
```
NottyCode, line 21: I would probably have said POM (Project Object Model) in this case. I think the general term is POM and to be honest I've been using maven for over a decade and I don't think I knew it even stood for that.

ref/general/packaging-deploying.adoc
```diff
@@ -14,79 +14,69 @@
...
 The link:https://github.com/OpenLiberty/ci.maven[Open Liberty Maven plug-in] helps manage your Open Liberty server and applications.
```
NottyCode, line 22: The phrase "helps manage your Open Libert server and application" could be read as meaning our plugin is for system management, rather than the plugin is there to help develop and package applications at the development phase.

ref/general/packaging-deploying.adoc
```diff
@@ -14,79 +14,69 @@
...
-
-After you enable the Maven plug-in, specify `war` in the `<packaging>` element in your `pom.xml` file.
-Then, run the `mvn package` command to package your application and Open Liberty into a WAR file.
-This command compiles, verifies, and builds the project.
-After you create this server package, you're ready to containerize and deploy it.
 To learn more about the Open Liberty Maven plug-in, check out this guide on link:/guides/maven-intro.html[Building a web application with Maven].
 
-=== Gradle
 Like Maven, link:https://gradle.org/[Gradle] is a build tool that provides dependency management for your projects.
 However, Gradle has it's own domain-specific language (DSL) that's based on Groovy.
```
NottyCode, line 26: I wouldn't do (DSL) in this case since it isn't reference again. If you are doing a compare and contrast I would probably mention somewhere that the pom file is XML and gradle builds are essentially a program. It is a key difference.

ref/general/packaging-deploying.adoc
```diff
@@ -14,79 +14,69 @@
...
-apply plugin: 'liberty'
-----
-Then, run the link:https://github.com/OpenLiberty/ci.gradle/blob/master/docs/libertyPackage.md#libertypackage-task[`libertyPackage` task] to package your application and Open Liberty into a WAR file.
-Once you create this server package, you can containerize and deploy it.
 To learn more about the Open Liberty Gradle plug-in, check out this guide on link:https://openliberty.io/guides/gradle-intro.html[Building a web application with Gradle].
 
-=== Runnable JAR file
-You can also use Maven or Gradle to package your application with Open Liberty into a runnable JAR file that contains the application and server runtime.
+== Server packages
+Open Liberty supports different server package types.
```
NottyCode, line 31: We haven't defined what a package is yet.

ref/general/packaging-deploying.adoc
```diff
@@ -14,79 +14,69 @@
...
+The supported package types are `zip`, `jar`, `tar`, and `tar.gz`.
+If you don't specify a package type, then `zip` is the default, meaning that your application and an Open Liberty installation are packaged into a ZIP file.
+
+For more information about Open Liberty packaging options with Maven, see the link:https://github.com/OpenLiberty/ci.maven/blob/master/docs/package.md#package[`package` goal in the Maven documentation]. For more information about Open Liberty packaging options with Gradle, see the link:https://github.com/OpenLiberty/ci.gradle/blob/master/docs/libertyPackage.md#libertypackage-task[`libertyPackage` task in the Gradle documentation].
+
+=== Packaging for containerization
+Use Maven or Gradle to package an application before you containerize it.
```
NottyCode, line 38: This use of package is different from the prior sections use of package as a term. In this case you are talking about creating a war and ear, in the prior one you were talking about a server package. I think this should probably exist prior to talking about server packages.

ref/general/packaging-deploying.adoc
```diff
@@ -14,79 +14,69 @@
...
+Maven and Gradle both work well as part of continuous integration and continuous deployment (CI/CD) pipelines.
+
+When you package an application for containerization and cloud deployment, specify `usr` in the `include` parameter or attribute.
+This option packages only relevant files in the `wlp.user.dir` directory, effectively excluding the runtime binary files from the compressed file.
+
+=== Packaging into a WAR deployment artifact
+You can also choose to package your application into a WAR deployment artifact for deployment on an Open Liberty server.
+Use Maven or Gradle to package your application into a WAR file so that it can be manually deployed on Open Liberty.
```
NottyCode, line 46: I think this should absolutely go first, this is basically what the pom will produce as the primary build artifact. The server package and containers are built on this primary function. This makes it sound like this is an option, rather than mandatory.

ref/general/packaging-deploying.adoc
```diff
@@ -14,79 +14,69 @@
...
+
+Specify `war` as the packaging type in your `pom.xml` file (for Maven) or in your `build.gradle` file (for Gradle). Then, package your application with Maven or Gradle. The result is a WAR deployment artifact that you can deploy on Open Liberty.
```
NottyCode, line 48: All uses of the package discussed are going to do this.

ref/general/packaging-deploying.adoc
```diff
@@ -14,79 +14,69 @@
...
+
+=== Packaging into a runnable JAR file
+If you want to package locally, you can package your application and an Open Liberty installation into a link:/docs/ref/general/#runnablejarfiles.html[runnable JAR file].
```
NottyCode, line 51: I'm not sure I understand the intent of the "if you want to package locally". How is this different from the other options?

ref/general/packaging-deploying.adoc
```diff
@@ -14,79 +14,69 @@
...
-----
-
-Each example overrides the `include` property to package your application and runtime into a runnable JAR file.
-The `minify` option specifies that the runnable JAR file contains only what you need to run your application so the application is as lightweight as possible.
-
-== Containerize
-After you create a server package by packaging your application with the Open Liberty runtime, you can containerize it for deployment.
+This local packaging method can be useful for debugging and testing during development.
+
+To package into a runnable JAR file, specify `runnable` as the package type in the `include` parameter or attribute.
```
NottyCode, line 55: This feels confused because the first sentence in the section says this is for creating runnable jars, but this sentence makes it sound like a runnable jar is a different option from what was being talked about in the first sentence.

ref/general/packaging-deploying.adoc
```diff
@@ -14,79 +14,69 @@
...
+You can also specify the `minify` option in the `include` property or attribute so that your runnable JAR file contains only what you need to run your application.
+
+== Containerization
+After you package your application into a server package with `usr` contents and build it with Maven or Gradle, you can containerize it for cloud deployment.
 Containerizing your server package bundles together the environment that your application needs to run.
 With Docker containers, you have the portability to run your application on any machine that is running Docker.
 
 A Docker image is a multi-layered file that serves as a template for a Docker container.
 The simplest way to containerize your server package is by using the latest link:https://hub.docker.com/_/open-liberty[Open Liberty Docker images], which include the most recent feature versions and link:/docs/ref/general/#security-vulnerabilities.html[security fixes].
```
NottyCode, line 64: I think Arthur would probably prefer we point people at the UBI images. In either case I think this calls for some kind of documentation on the different images  and how to decide which ones to use.

ref/general/packaging-deploying.adoc
```diff
@@ -14,79 +14,69 @@
...
 Finally, the `docker run` command runs your container that houses the application.
 
-To learn more about running applications with Open Liberty Docker images, see these guides on link:https://openliberty.io/guides/docker.html[Using Docker containers to develop microservices] and link:/guides/containerize.html[Containerizing microservices].
+To learn more about running applications with Open Liberty Docker images, see this guides on link:/guides/containerize.html[Containerizing microservices].
 
-== Deploy
-After you containerize your server package, you can deploy it to link:https://kubernetes.io/[Kubernetes] or link:https://www.openshift.com/[OpenShift].
+== Deployment
+You can deploy web applications and enterprise applications on Open Liberty.
+Deploy containerized applications to the cloud, or, for traditional environments, deploy a WAR deployment artifact on Open Liberty.
```
NottyCode, line 76: I'm having trouble with this sentence. It reads like an instruction, but then there aren't any actual instructions.

ref/general/packaging-deploying.adoc
Charlotte-Holt: I updated the link. Yeah, I think that would be outside the scope of this topic, but we probably need to add something to the backlog.

PR #1438, opened 2020-04-17
Updates per David's peer review

Charlotte-Holt:


ref/general/securityUtility.adoc
```diff
@@ -205,13 +205,13 @@ The signature algorithms that end with RSA create certificates with RSA keys.
 The signature algorithms that end with ECDSA create certificates with Elliptical Curve (EC) keys.
 {empty} +
 {empty} +
-If you use certificates that are created with EC keys, then include a customized ciphers list in your TLS configuration to include EC ciphers.
+If you use certificates that are created with EC keys, then include a customized ciphers list in your TLS configuration that includes EC ciphers.
```
dmuelle, line 208: looking at this again it should probably be

If you use certificates that are created with EC keys, then include a customized ciphers list that includes EC ciphers in your TLS configuration.

ref/general/securityUtility.adoc
```diff
@@ -52,26 +51,27 @@ securityUtility action [options]
 ----
 
 action::
-The action for the tool to perform.
+The action for the command to perform.
```
dmuelle, line 54: should this be

The action that the command performs.

PR #1455, opened 2020-04-23
Update mpJwt example doc

ManasiGandhi:


ref/feature/mpJwt/examples.adoc
```diff
@@ -1,3 +1,5 @@
+== Examples
+
```
dmuelle, line 2: Is the feature manager config required for this example? It's not usually included since the feature doc already has a line to add the feature. Is the current intro sentence needed?

ref/feature/mpJwt/examples.adoc
```diff
@@ -13,23 +15,40 @@ The `<mpJwt/>` element is required to configure the processing of the caller’s
         <feature>cdi-2.0</feature>
         <feature>appSecurity-3.0</feature>
     </featureManager>
+----
 
+This is the keystore that is used by SSL and by JWT.
```
dmuelle, line 20: it might be helpful to combine the code into a single block, with an explanation beforehand, as this is how most of the other config snippets are formatted- unless this one needs to be different. You might also want a subheading for the unified example, like "Basic configuration to implement MicroProfile JWT"

ref/feature/mpJwt/examples.adoc
```diff
@@ -13,23 +15,40 @@ The `<mpJwt/>` element is required to configure the processing of the caller’s
...
+The keystore is built by using the maven keytool plugin,
 
-    <!-- This is the keystore that will be used by SSL and by JWT.
-         The keystore is built using the maven keytool plugin -->
     <keyStore id="defaultKeyStore" location="keystore.jceks" type="JCEKS"
               password="secret" />
 
-    <!-- The HTTP ports that the application will use. -->
+These are the HTTP ports that the application uses,
```
dmuelle, line 26: don't use "These" as the subject- could be rewritten in "The following example shows..." format

ref/feature/mpJwt/examples.adoc
```diff
@@ -1,35 +1,33 @@
...
 </server>
 ----
+
+=== Specify login with Micrprofile JSON Web Token for only a subset of applications, URLs, browsers, or IP addresses
+
+To protect only a subset of applications, URLs, or IP addresses, use an authentication filter. The security configuration takes effect only when the conditions in the filter are met. In the following example, requests that target the `myApp` application are processed by the mp-jwt feature:
+
+[source,xml]
+----
+<mp-jwt authFilterRef="authFilter1" />
```
dmuelle, line 27: does `mp-jwt` require additional config here, as in the above example? I'm not sure

ref/feature/mpJwt/examples.adoc
```diff
@@ -1,35 +1,33 @@
-The `mpJwt` feature adds the libraries that are required for MicroProfile JWT implementation.
+== Examples
+
+=== Basic configuration to implement MicroProfile JWT
 
 The `<mpJwt/>` element is required to configure the processing of the caller’s JWT.
 
+The Secure Sockets Layer(SSL)and the JSON Web Token(JWT) uses a keystore by using the maven keytool plugin.The keystore is built by using the maven keytool plugin.The applicaton uses HTTP ports and contains the user and login endpoints. The MP JWT configuration injects the caller's JWT into a ResourceScoped bean for inspection.
```
dmuelle, line 7: `The Secure Sockets Layer(SSL)and the JSON Web Token(JWT) uses a keystore by using the maven keytool plugin.`
- add space between ) & and
- "uses" should be "use" since subject is plural
- I'm not sure what this means- SSL is a protocol, so it doesn't necessarily "use" anything- Not sure it needs to be explained in this context but the relationship between keystore and keytool plugin is not clear.  Does JWt access the keystore by using the plugin?
- need a space between period and "The"  `plugin.The keystore`

`The keystore is built by using the maven keytool plugin.`

Does this mean- The Maven keytool plugin creates the keystore if one does not already exist? might be a question for SME- more specifically- why does the keystore matter to JWT config?


`The applicaton uses HTTP ports and contains the user and login endpoints.`
- application (sp)
- I'm not sure what this means. It might be helpful to call out the elements (eg. `httpEndpoint`) in the description so the user can easily connect them to the config example

Are the Keystore, webapp, http endpoint, etc part of the MPjwt config?  They are explained but not put in context- and the attributes of the mp-jwt element aren't explained

PR #1459, opened 2020-04-23
Update docs to indicate Java 14 is supported

aguibert:


NottyCode:
What about including a link to where to download Java SE 14 from AdoptOpenJDK so if someone gets here and thinks they want a JVM they have somewhere to go?

The PR should go to develop not master.

PR #1531, opened 2020-05-07
Updates per David's edit suggestions

ManasiGandhi:


ref/general/microshed.adoc
```diff
@@ -6,37 +6,35 @@
...
-== Using MicroShed Testing
+== Writing integration tests with Microshed Testing
 
-You can easily deploy your microservices in different environments by using containers.
-Containers are convenient and combine everything that you need to run a service into a single unit.
-They are lightweight, disposable, configurable, and manageable with an orchestration software like Kubernetes.
-You can leverage containers in a test environment by using the Testcontainers framework, which is an open source library that supports JUnit tests.
+You can deploy your microservices in different environments by using containers.
+With the Testcontainers framework, which is an open source library that supports JUnit tests, you can use containers in a test environment.
+Microshed implements testcontainers to support dev/prod parity in testing.
```
dmuelle, line 37: Microshed implements testcontainers to support dev/prod parity in testing.

Should this be "Microshed Testing" ?and should Testcontainers be capitalized?

I think if you reversed the order of sentences on lines 37 and 38 it might flow better. That way you would introduce microshed before explaining it's relation to testcontainers

PR #1553, opened 2020-05-13
Add section to securityUtility doc

Charlotte-Holt:


ref/general/securityUtility.adoc
```diff
@@ -214,3 +215,41 @@ If the `--extInfo` option is used, the value is passed directly to the link:http
...
+
+[#limits-to-encryption]
+== Limits to protection through password encryption
+
+Open Liberty supports AES encryption for passwords that are stored in the `server.xml` file.
+When you use this option for protecting system passwords in the Open Liberty configuration, you need to understand the limits to the protection it provides.
+
+Encrypting a password in the Open Liberty configuration doesn't guarantee that the password is secure or protected.
+It only means that someone who can see the encrypted password but doesn't know the encryption key, cannot easily recover the password.
+The application server process requires access to both the encrypted password and the decryption key, so both these items need to be stored on the file system that is accessible to the server runtime environment.
```
ManasiGandhi, line 229: "The application server process requires access to both the encrypted password and the decryption key, so both these items need to be stored on the file system that is accessible to the server runtime environment." Here, can you be specific about what the application server process is for? For example, authentication etc.

Issue #1462, opened 2020-04-27
Example config for JPA Container feature


Charlotte-Holt:
Example config for the JPA Container feature: https://openliberty.io/docs/ref/feature/#jpaContainer-2.2.html

dmuelle:
comment from @ggam in the PR used to create initial draft: (https://github.com/OpenLiberty/docs/pull/1579)

" I just used this feature today. In my tests and after reading the WAS Liberty doc, I came to the conclusion that the application classloader needs to see the Hibernate lib for this to work (btw the example here doesn't reference the lib so I don't think it will work).

I also miss an example using bell."

dmuelle:
First draft now in preview at https://draft-openlibertyio.mybluemix.net/docs/ref/feature/#jpaContainer-2.2.html

This example was taken from @frowe code snippet in #1291 

Further examples pending SME review

dmuelle:
Additional resources:

https://stackoverflow.com/questions/47396310/open-liberty-and-hibernate

https://www.ibm.com/support/knowledgecenter/SSEQTP_liberty/com.ibm.websphere.wlp.doc/ae/twlp_dep_jpa.html

dmuelle:
Hi @frowe can you review this example (taken from your comment in #1291 ) and let me know if further examples are needed. See the user comment above re an example with `bells-1.0` . If there's anyone else I should get to review this, just let me know.

dmuelle:
Hi @frowe - circling back on this. Can you let me know by Wednesday 6/17 if the example is correct and if further examples are needed for this topic? We are targeting this for 20.0.0.7 if possible. Thanks!

frowe:
@dmuelle We need to extend the current config snippet to include configuration of the app class loader as seen in https://www.ibm.com/support/knowledgecenter/en/SSD28V_liberty/com.ibm.websphere.wlp.core.doc/ae/twlp_dep_jpa.html.  For an application named "myApp", you would use:

```
<jpa defaultPersistenceProvider="org.hibernate.jpa.HibernatePersistenceProvider" />
<library id="HibernateLib">
    <fileset dir="${shared.resource.dir}/jpa21_hibernate" includes="*.jar"/>
</library>
<application location="myApp.war">
   <classloader commonLibraryRef="HibernateLib"/>
</application>
```

dmuelle:
[...]

PR #1579, opened 2020-05-19
create examples and includes

dmuelle:


ggam:
I just used this feature today. In my tests and after reading the WAS Liberty doc, I came to the conclusion that the application classloader needs to see the Hibernate lib for this to work (btw the example here doesn't reference the lib so I don't think it will work).

I also miss an example using bell.

dmuelle:
thank you @ggam - this doc is still in development and this is just an initial draft to get SME feedback. I will copy your comments into the issue where we are tracking this doc: #1462

PR #1581, opened 2020-05-20
Updates per LC's review

Charlotte-Holt:


ref/general/password-encryption.adoc
```diff
@@ -17,21 +17,26 @@ Open Liberty supports AES encryption for passwords that are stored in the `serve
...
 The application server process requires access to both the encrypted password and the decryption key, so both these items need to be stored on the file system that is accessible to the server runtime environment.
 The encryption key is also required by anyone who encrypts a password that is placed in the server configuration.
-For an attacker that has access to exactly the same set of files as the Open Liberty server instance, applying AES encryption to the password provides no additional security over XOR encoding.
+For an attacker that has access to the same set of files as the Open Liberty server instance, applying AES encryption to the password provides no additional security over XOR encoding.
 
 However, there are still reasons why you might consider encrypting passwords in the Open Liberty configuration.
-The Open Liberty configuration is designed to be highly composable and sharable.
-As such, it’s designed to be configured without an administration subsystem, so any XOR-encoded password is visible to any administrator.
-Given these design features, consider the following scenarios:
+The Open Liberty configuration is highly composable and sharable.
```
Rwalls1, line 30: I think you could remove the word "highly" here as it may not add value to the statement without additional context. In general, I think it is usually best to avoid qualitative statements that may not be verifiable without a point of comparison or additional information.

ref/general/password-encryption.adoc
```diff
@@ -17,21 +17,26 @@ Open Liberty supports AES encryption for passwords that are stored in the `serve
...
+As such, it can be configured without an administration subsystem, so any XOR-encoded password is visible to any administrator.
```
Rwalls1, line 31: I think you could remove "As such" here as I think the sentence doesn't lose its meaning without its inclusion.

PR #1836, opened 2020-07-06
More updates, requesttTming

ManasiGandhi:


ref/general/requestTiming.adoc
```diff
@@ -22,6 +22,12 @@ For a hung request, apart from printing a warning message along with the relevan
 Three Java cores are generated for a hung request, each a minute apart.
 If a previously hung request is completed, a detailed message is logged, including the ID and time of completion of the request.
 
+A slow request can be slow because the code path length of a request is particularly long or because some resource the request depends on, such as a database is busy.
```
dmuelle, line 25: I think you need a comma after "database"

ref/general/requestTiming.adoc
```diff
@@ -22,6 +22,12 @@ For a hung request, apart from printing a warning message along with the relevan
...
+Requests that take much longer, and which are therefore likely to be hung, are typically blocked by lack of access to a shared resource, for example, a section of code that only allows serial access and transfers information sequentially.
```
dmuelle, line 26: this sentence is too long for acrolinx. I think you could break it in two without changing much. Also the "only" should be before "serial".

ref/general/requestTiming.adoc
```diff
@@ -22,6 +22,12 @@ For a hung request, apart from printing a warning message along with the relevan
...
+In the case of a hung request, this can be blocked for a long time or even indefinitely.
```
dmuelle, line 27: Avoid using "this" as subject of clause: 

`A hung request can be blocked for a long time or even indefinitely.`

ref/general/requestTiming.adoc
```diff
@@ -22,6 +22,12 @@ For a hung request, apart from printing a warning message along with the relevan
...
+One particular form of hung request occurs when two threads are blocked waiting on each other to return shared resources, which is called a deadlock.
+For a slow request we need to know what actions have occurred during the processing of that request, and where in the request the processing is currently at.
```
dmuelle, line 29: just a thought- this sentence might work better directly after  the slow request description, before you talk about hung requests

Also better to avoid "we" as acrolinx will flag it

PR #1895, opened 2020-07-08
Edits per David's peer edit suggestions

ManasiGandhi:


ref/feature/samlWeb/examples.adoc
```diff
@@ -1,26 +1,26 @@
...
 This default configuration of the feature activates the following endpoints:
 
-- AssertionConsumerService URL:
-The AssertionConsumerService URL is the endpoint on the service provider point of contact server that receives assertions.
-
+- The AssertionConsumerService URL is the endpoint on the service provider point of contact server that receives assertions.
++
 `https://<hostname>:<sslport>/ibm/saml20/defaultSP/acs`
 
 - Service Provider (SP) metadata URL is used to provide configuration information for the  service provider
```
dmuelle, line 12: I think you need "The" here, as in the previous bullet

ref/feature/samlWeb/examples.adoc
```diff
@@ -1,26 +1,26 @@
...
-You can use a browser to download the metadata for this SP by using this URL and provide the URL to the SAML IdP to establish federation between this SP and IdP.
+You can use a browser to download the metadata for this SP from this URL. Provide the URL to the SAML IdP to establish federation between this SP and IdP.
 
 === Custom configuration of the SAML web SSO
 
-In most cases, exchange of the metadata files is sufficient.
-Additional attributes can be specified if necessary.
+In most cases, exchange of the metadata files is sufficient for verification.
+You can specify additional attributes if necessary.
+For more information, see the feature configuration elements.
```
dmuelle, line 23: I think if you put the last two sentences before the example in a new paragraph, it would make it easier to read in the context of the example. As it is, its a little hard to tell that the example really only pertains to the last two sentences

Issue #1932, opened 2020-07-14
Replace "config" with configuration 


dmuelle:
"Config" is not recognized by IBM terminology as an abbreviation. Replace all instances of "config", unless part of a product or feature name (eg MP Config) or code example.

particularly in [Server configuration overview](https://www.openliberty.io/docs/ref/config/),  but check through all the docs.

dmuelle:
https://draft-openlibertyio.mybluemix.net/docs/20.0.0.9/reference/config/server-configuration-overview.html

The server configuration overview was the only doc that used this abbreviation- searched all the published docs and didi not find another instance. The updates will go to production with 20.0.0.09, leaving tis bug open until they are confirmed.

dmuelle:
fix delivered to vNext in #2361

PR #2198, opened 2020-08-03
2101 staging changes

dmuelle:


dmuelle:
Also contains changes per #1932

PR #2531, opened 2020-09-01
Improve the doc linking macro so that display text can be shown.

steven1046:
The docs macro works right now but it only shows the filename (without extension) on the UI. This will improve it so that we can specify the display text in the macro links in our .adoc files. E.g. `feature:appSecurity[display=Application Security feature]` will show up as `Application Security feature` now such as in the second paragraph:
image

### DIFF
--- a/lib/ol-asciidoc.js
+++ b/lib/ol-asciidoc.js
@@ -3,7 +3,7 @@ const { posix: path } = require('path');
 function DocLinkMacros (context, docType) {
   return function () {
     this.process((parent, target, attrs) => {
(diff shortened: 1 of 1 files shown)


steven1046:
This fixes the display text part of https://github.com/OpenLiberty/openliberty.io/issues/1755.

steven1046:
I asked in the asciidoctor Gitter channel how we should implement this, and they said we would have to used a named asciidoc attribute to make it work.

ellenwyllie (approved):

PR #2981, opened 2020-10-26
XSD nav.doc fle update

ManasiGandhi:


Charlotte-Holt:
The `nav.adoc` files should only be updated in the `draft-nav` and `staging-nav` branches.

PR #3796, opened 2021-03-10
Update external-configuration.adoc

ManasiGandhi:


Joseph-Cass (approved):
Approving so we can see the generated draft, but I don't think this is ready for a final version yet.

PR #3972, opened 2021-04-09
Update the Security vulnerability for OL jsf

clarkek123:
This update is being made to limit the JSF  feature levels to those which are in OL.


dmuelle (approved):
LGTM

PR #4045, opened 2021-04-14
3801 jwtsso

ManasiGandhi:


modules/reference/pages/feature/appSecurity/examples.adoc
```diff
@@ -49,11 +49,12 @@ When the Application Security feature is enabled, Lightweight Third Party Authen
 
 LTPA configuration can provide Single Sign-on (SSO) to secure applications. For more information, see link:/docs/ref/general/#single-sign-on.html[Single sign-on (SSO)].
 
-=== Configure an authentication filter
+=== Configure an authentication filter for JWT LTPA authentication
```
dmuelle: does this example relate to jwt?

The titles still don't indicate the value of the filter- which is to specify whether requests are processed with a particular mechanism (in this case LTPA). 
For example, you wouldn't call an article that describes how to fix a leaky pipe "Use a wrench".  The filter is just the tool

dmuelle (approved):

PR #4229, opened 2021-05-20
clarify separation of batch vs admin roles

scottkurz:
Signed-off-by: Scott Kurz <skurz@us.ibm.com>

Sorry I didn't catch this new article before the initial effort and am only commenting now.

@dmuelle is it OK to link to the WL  KCtr (whatever it's called now) from the OL doc?  

If so, I might reference:   https://www.ibm.com/docs/en/was-liberty/nd?topic=liberty-securing-batch-environment

Note the first two changes here removed a non-display char that I assume was added by mistake.

### DIFF
--- a/modules/ROOT/pages/admin-center.adoc
+++ b/modules/ROOT/pages/admin-center.adoc
@@ -136,24 +136,25 @@ Each batch job has an **Actions** icon, which you can select to stop, restart, o
 
 If batch jobs or job logs are on remote servers, link:/guides/cors.html[configure cross origin region sharing (CORS)] on each remote server. CORS enables Admin Center to request job information from remote servers.
 
-Note too that the batch features use a custom authorization specific to batch and role membership must be granted separately from the roles needed to access the Admin Center in the first place.  
-
-E.g. a 'batchadmin' user configured like:
+The Batch Management feature requires custom authorization to view and manage batch jobs. To use the Java Batch tool, you must configure a `com.ibm.ws.batch` custom authorization role, in addition to the reader or administrator role that is required to access the Admin Center. The following server.xml example shows configuration for a user, `wanda`, who is granted the administrator management role and the batchAdmin custom authorization role:
 
 [source,xml]
 ----
-  <reader-role>
-      <user>batchadmin</user>
-  </reader-role>
+  <administrator-role>
+      <user>wanda</user>
+  </administrator-role>
 
   <authorization-roles id="com.ibm.ws.batch">
       <security-role name="batchAdmin">
-          <user name="batchadmin"/>
+          <user name="wanda"/>
       </security-role>
   </authorization-roles>
 ----
 
-could use the Java Batch tool to view and operate on any and all jobs (as 'batchAdmin') but a user with just 'administrator-role' access might require one of the 'com.ibm.ws.batch' custom roles in order to even view any jobs at all (they are not automatically a 'batch-admin').
+With this configuration, the user has authorization to view and manage any configured Java batch jobs.  The `batchAdmin` authority could be combined with a more fine-grained, restrictive management authority by only granting 'wanda' the `reader-role` management role, which would still allow access to the Admin Center Java Batch tool.
+
+On the other hand, if no custom batch authorization role is configured, even a user in the administrator management role cannot view or manage Java batch jobs. For more information, see https://www.ibm.com/docs/en/was-liberty/nd?topic=liberty-securing-batch-environment[Securing the Liberty batch environment].
+
 
 [#openid]
 === Administer Open ID Connect Provider tasks with the OpenID Connect (OIDC) tools
--- a/modules/ROOT/pages/admin-center.adoc
+++ b/modules/ROOT/pages/admin-center.adoc
@@ -151,9 +151,11 @@ The Batch Management feature requires custom authorization to view and manage ba
   </authorization-roles>
 ----
 
-With this configuration, the user has authorization to view and manage any configured Java batch jobs.  The `batchAdmin` authority could be combined with a more fine-grained, restrictive management authority by only granting 'wanda' the `reader-role` management role, which would still allow access to the Admin Center Java Batch tool.
+With this configuration, the user has authorization to view and manage any configured Java batch jobs.
 
-On the other hand, if no custom batch authorization role is configured, even a user in the administrator management role cannot view or manage Java batch jobs. For more information, see https://www.ibm.com/docs/en/was-liberty/nd?topic=liberty-securing-batch-environment[Securing the Liberty batch environment].
+The `batchAdmin` role can also be combined with a more restrictive Liberty management role by assigning 'wanda' to `reader-role`. This combination still allows full access to the Java Batch tool but provides read-only access to other Admin Center resources.
+
+However, if no custom batch authorization role is configured, even a user in the administrator management role cannot view or manage Java batch jobs. For more information, see https://www.ibm.com/docs/en/was-liberty/nd?topic=liberty-securing-batch-environment[Securing the Liberty batch environment].
 
 
 [#openid]
--- a/modules/ROOT/pages/admin-center.adoc
+++ b/modules/ROOT/pages/admin-center.adoc
@@ -67,7 +67,7 @@ image::ui_login.png[The Admin Center login screen,align="center"]
 
 == Select tools from the Toolbox
 
-After you log in toÂ Admin Center, the browser displays the **Toolbox**, which contains tools such as the **Server Config** and **Explore** tools and a bookmark to link:https://openliberty.io[openliberty.io]. The following screen capture shows the Admin Center **Toolbox**:
+After you log in to Admin Center, the browser displays the **Toolbox**, which contains tools such as the **Server Config** and **Explore** tools and a bookmark to link:https://openliberty.io[openliberty.io]. The following screen capture shows the Admin Center **Toolbox**:
 
 image::ui_toolbox.png[align="center"]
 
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/admin-center.adoc
```diff
@@ -128,14 +128,33 @@ You can add more resource metrics to the **Monitor** view by selecting the **Edi
...
+If you configure�the feature:batchManagement[display=Batch Management] feature, you can access the **Java Batch** tool. With this tool, you can view the progress and status of your Java batch jobs, manage their instances, and view their log files. The following screen capture shows the **Java Batch** tool:
 
 image::ui_javaBatchTool.png[align="center"]
 
 Each batch job has an **Actions** icon, which you can select to stop, restart, or purge the job, and a **View Logs** icon, which you can select to view the job logs. To view details and metrics for a batch job, hover over the **Batch Job Name** and select an option from the tooltip menu.
 
 If batch jobs or job logs are on remote servers, link:/guides/cors.html[configure cross origin region sharing (CORS)] on each remote server. CORS enables Admin Center to request job information from remote servers.
 
+Note too that the batch features use a custom authorization specific to batch and role membership must be granted separately from the roles needed to access the Admin Center in the first place.
```
dmuelle: Suggested edit:

The Batch Management feature requires custom authorization to view and manage batch jobs. To use the Java Batch tool, you must configure a `com.ibm.ws.batch` custom authorization role, in addition to the reader or administrator role that is required to access the Admin Center. The following `server.xml` example shows configuration for a `batchadmin` user who is granted the `reader-role` management role and the `batchAdmin` custom authorization role: 

[example]

With this configuration, the user has authorization to view and manage any configured Java batch jobs. If no custom authorization role is configured, even a user in the administrator management role cannot view or manage Java batch jobs. For more information, see https://www.ibm.com/docs/en/was-liberty/nd?topic=liberty-securing-batch-environment[Securing the Liberty batch environment].

modules/ROOT/pages/admin-center.adoc
```diff
@@ -128,14 +128,33 @@ You can add more resource metrics to the **Monitor** view by selecting the **Edi
...
 
 If batch jobs or job logs are on remote servers, link:/guides/cors.html[configure cross origin region sharing (CORS)] on each remote server. CORS enables Admin Center to request job information from remote servers.
 
+Note too that the batch features use a custom authorization specific to batch and role membership must be granted separately from the roles needed to access the Admin Center in the first place.  
+
+E.g. a 'batchadmin' user configured like:
+
+[source,xml]
+----
+  <reader-role>
```
dmuelle: i think it's potentially confusing in this example to have one role called `batchadmin` and another called `batchAdmin`. If the security-role requires `batchAdmin` can we rename the reader role to something like `batchToolUser` or `wanda`?

dmuelle (changes requested):
Hi @scottkurz thanks for opening this- can you retarget it to `draft`? We'll need to stage it there and then on `staging` before the change goes to vNext.

As to your question re linking to IBM docs- we don't generally link from OL to KC/IBM Docs. I think it's ok to do so temporarily in this case if that's the only source for more information.

A better long term solution is to develop a Java batch topic for the OL docs that covers all aspects of the KC info that are relevant to Open Liberty users. If you wouldn't mind opening an issue on the docs backlog to help us understand what topics/info from IBM docs would make sense in an OL Java batch topic, I can get it prioritized and work up a draft. Let me know what you think. Thanks!

modules/ROOT/pages/admin-center.adoc
```diff
@@ -128,14 +128,33 @@ You can add more resource metrics to the **Monitor** view by selecting the **Edi
...
+      <user>batchadmin</user>
+  </reader-role>
+
+  <authorization-roles id="com.ibm.ws.batch">
+      <security-role name="batchAdmin">
+          <user name="batchadmin"/>
+      </security-role>
+  </authorization-roles>
+----
```
dmuelle, line 152: also want to clarify- even a user in the reader role can edit batch jobs if they have the proper `com.ibm.ws.batch` auth? I thought that reader role gave you read-only access to the Admin center tools- is that not the case with Java Batch?

scottkurz: Yeah, this was exactly the nuance I was trying to illustrate.    I think it makes sense to basically defer to batch's custom auth here but let me also tag @mbroz2  to clarify if this is truly the intended design.

modules/ROOT/pages/admin-center.adoc
scottkurz: It's one user, batchadmin, added to each of two different roles:  reader-role and batchAdmin.  The user needs both roles.

modules/ROOT/pages/admin-center.adoc
scottkurz: That looks good (and better) to me.   Might as well wait for the other comments before going through with the changes.

scottkurz:
> A better long term solution is to develop a Java batch topic for the OL docs that covers all aspects of the KC info that are relevant to Open Liberty users. 

Hmm...the challenge is it's potentially all of them https://www.ibm.com/docs/en/was-liberty/nd?topic=liberty-deploying-java-batch-applications-in

I imagine we don't want literally of that but not sure where to start.   Shall I open an issue at least to start the conversation?

modules/ROOT/pages/admin-center.adoc
dmuelle: Yes, I know its the same user, but i think potential for confusion is between the user name `batchadmin` and the role name `batchAdmin`. I was just thinking if either of those example IDs is arbitrary, then making them a little more different would help disambiguate their purpose in the example. 

As is, a reader might not catch the minor difference in the two strings and wonder if they need to match the role to the ID in their environment (Spoiler alert: this happened to me :)

Not a major concern- if the config requires both names this way, or if you just disagree, I'm ok with it as is.

dmuelle:
> Hmm...the challenge is it's potentially all of them https://www.ibm.com/docs/en/was-liberty/nd?topic=liberty-deploying-java-batch-applications-in
> 
> I imagine we don't want literally of that but not sure where to start. Shall I open an issue at least to start the conversation?

yes please! We can start there and determine what we need and dont need, or if the organization/format would benefit from revision. I tend to think we'd be better off with one or two longer, well organized topics with internal anchor links vs a bunch of short topics as it is in KC

modules/ROOT/pages/admin-center.adoc
mbroz2: reader-role is not an Admin Center concept, but rather a Liberty one (aka, it's applied to all (well almost all) APIs, including Admin Center ones, unless the API uses their own set of authorization, like batch).  
I would expect a user to be be able to have the 'reader' role and the 'batchAdmin' role, allowing them to view all the tools in admin center, but also perform batch admin job operations in the java batch tool.... but that's only true if no other/non-java-batch APIs are called that would do anything other then a GET operation (except for the Admin Center user preferences API, which is allowed to do POST/PUT/DELETE for reader role).  I'm guessing there aren't, but @kinueng or @k8vance88 could confirm.  One way to test this would be to just run the Admin Center Java Batch FAT tests, and change the user's role from (what I'm guessing is) admin to reader.

Of course the cheapest route is to simply change the example to asign the batchadmin user to the admin role instead of the reader role :)

modules/ROOT/pages/admin-center.adoc
scottkurz: That looks good (and better) to me. Might as well wait for the other comments before going through with the changes.

modules/ROOT/pages/admin-center.adoc
scottkurz: It's one user, batchadmin, added to each of two different roles: reader-role and batchAdmin. The user needs both roles.

modules/ROOT/pages/admin-center.adoc
scottkurz: Yeah, this was exactly the nuance I was trying to illustrate. I think it makes sense to basically defer to batch's custom auth here but let me also tag @mbroz2 to clarify if this is truly the intended design.

scottkurz:
> yes please! We can start there and determine what we need and dont need, or if the organization/format would benefit from revision. I tend to think we'd be better off with one or two longer, well organized topics with internal anchor links vs a bunch of short topics as it is in KC

Opened https://github.com/OpenLiberty/docs/issues/4231

modules/ROOT/pages/admin-center.adoc
dmuelle: I understand that, the question I had was whether it might be confusing to readers to arbitrarily use two near-identical strings (`batchadmin` & `batchAdmin`) for two different example values in the same code sample.

dmuelle:
> Opened #4231
 thank you!

modules/ROOT/pages/admin-center.adoc
scottkurz: Thx @mbroz2.   Yeah, if we just want the sample snippet to help a single developer copy/paste working config we probably just want to use the admin role.   I don't think you need 'admin' for anything purely batch related but it can be useful e.g. for  **ibm/api/validation**.

I was trying to capture the angle "what is the minimum auth required?" for users trying to use the roles in a fine-grained manner.   

Maybe a good compromise would be to show the sample snippet with admin role but mention in a sentence that you can use reader-role for batch operations.

modules/ROOT/pages/admin-center.adoc
scottkurz: OK, sure, good point.     Sorry.. I wasn't trying to be stubborn I just somehow left my comment twice :)

modules/ROOT/pages/admin-center.adoc
```diff
@@ -136,24 +136,25 @@ Each batch job has an **Actions** icon, which you can select to stop, restart, o
...
-          <user name="batchadmin"/>
+          <user name="wanda"/>
       </security-role>
   </authorization-roles>
 ----
 
-could use the Java Batch tool to view and operate on any and all jobs (as 'batchAdmin') but a user with just 'administrator-role' access might require one of the 'com.ibm.ws.batch' custom roles in order to even view any jobs at all (they are not automatically a 'batch-admin').
+With this configuration, the user has authorization to view and manage any configured Java batch jobs.  The `batchAdmin` authority could be combined with a more fine-grained, restrictive management authority by only granting 'wanda' the `reader-role` management role, which would still allow access to the Admin Center Java Batch tool.
+
+On the other hand, if no custom batch authorization role is configured, even a user in the administrator management role cannot view or manage Java batch jobs. For more information, see https://www.ibm.com/docs/en/was-liberty/nd?topic=liberty-securing-batch-environment[Securing the Liberty batch environment].
```
dmuelle: On the other hand
--->
However

modules/ROOT/pages/admin-center.adoc
```diff
@@ -136,24 +136,25 @@ Each batch job has an **Actions** icon, which you can select to stop, restart, o
...
   <authorization-roles id="com.ibm.ws.batch">
       <security-role name="batchAdmin">
-          <user name="batchadmin"/>
+          <user name="wanda"/>
       </security-role>
   </authorization-roles>
 ----
 
-could use the Java Batch tool to view and operate on any and all jobs (as 'batchAdmin') but a user with just 'administrator-role' access might require one of the 'com.ibm.ws.batch' custom roles in order to even view any jobs at all (they are not automatically a 'batch-admin').
+With this configuration, the user has authorization to view and manage any configured Java batch jobs.  The `batchAdmin` authority could be combined with a more fine-grained, restrictive management authority by only granting 'wanda' the `reader-role` management role, which would still allow access to the Admin Center Java Batch tool.
```
dmuelle: suggested edit:
With this configuration, the user has authorization to view and manage any configured Java batch jobs. 

The `batchAdmin` authority can also be combined with a more restrictive management authority by granting 'wanda' the `reader-role` management role. This combination still allows full access to the Java Batch tool but provides read-only access to other Admin Center resources.

dmuelle (changes requested):
Looks good- just a minor suggestion for ID guidance nitpicks

modules/ROOT/pages/admin-center.adoc
mbroz2: personally, I would change 
"The batchAdmin authority can also be combined with a more restrictive management authority by granting 'wanda' the reader-role management role." 
to 
"The `batchAdmin` role can also be combined with a more restrictive Liberty management role by assigning 'wanda' to `reader-role`" 
as I think that reads easier and avoids 'authority'

scottkurz: Thanks this wording didn't come easy to me so appreciate the input.   I combined the comments and pushed up take # 3.

dmuelle (approved):
LGTM, thanks!

PR #4246, opened 2021-05-25
Update distributed-session-caching.adoc

ManasiGandhi:


modules/ROOT/pages/distributed-session-caching.adoc
```diff
@@ -55,9 +55,10 @@ Client-Server model:: An Open Liberty server can act as the JCache client and co
 
 Peer-to-Peer model:: An Open Liberty server can connect with other Open Liberty servers that are also running with the JCache Session Persistence feature and configured to be part of the same cluster.
 
-The examples show two applications, service A and service B, that are both running on different Open Liberty servers that are part of the same cluster.
+The following examples show two applications, service A and service B, that are both running on different Open Liberty servers that are part of the same cluster.
 In the example, service A sets the value for the `foo` attribute to equal `bar`.
 Service B can then retrieve that value for the `foo` attribute from the distributed cache that is shared by all servers in the cluster.
+Your choice of model depends on the needs of your application, the resources that are available, and the specifications of your chosen Cache implementation.
```
dmuelle: your chosen Cache implementation
--->
your chosen JCache implementation

modules/ROOT/pages/distributed-session-caching.adoc
```diff
@@ -88,10 +89,13 @@ public class ServiceB extends HttpServlet {
...
 
 Each JCache implementation configures the cluster instances differently.
-See the documentation for your chosen implementation.
+For more information, see the documentation for your chosen implementation.
 
-You don’t need to specify a Hazelcast configuration file if the default Hazelcast configuration works for you.
-However, you can create configuration file similar to the following example and add a reference to it from your `server.xml` file to customize how Hazelcast defines the cluster:
+== Customize your JCache implementation
+
+Most JCache implementations provide default settings.
```
dmuelle: I think it's likely that all implementations provide default settings

I think you should say something like:
`Most JCache implementations provide sensible default settings.`

doesn't have to be "sensible", though there is precedent for that term. vis-a-vis "[convention over configuration](https://en.wikipedia.org/wiki/Convention_over_configuration)"

The idea to convey is that in many cases you don't need to customize the settings, so users know this is not necessarily required

modules/ROOT/pages/distributed-session-caching.adoc
ManasiGandhi: updated.

modules/ROOT/pages/distributed-session-caching.adoc
ManasiGandhi: updated.

dmuelle (approved):

PR #4409, opened 2021-07-01
2408 xml schema 3

ManasiGandhi:


modules/reference/pages/xsd-schema.adoc
```diff
@@ -0,0 +1,60 @@
...
+//
+:page-description: Open Liberty includes tools for creating an XSD schema for the `server.xml` file.
+:page-layout: general-reference
+:seo-title: XSD schema for the server.xml file
+:seo-description: Open Liberty includes tools for creating an XSD schema for the `server.xml` file.
+:page-type: general
+= XML schema for the server.xml file
+
+The XML Schema Definition (XSD) language is a markup language that defines the structure and data types of XML documents.
+Open Liberty includes tools that create an XSD file that generates a schema and validates the `server.xml` file of your application.
```
dmuelle: Open Liberty includes tools that create an XSD file that generates a schema and validates the `server.xml` file of your application.
--->
Open Liberty includes tools to create an XSD schema file that you can use  to validate your `server.xml` file.

modules/reference/pages/xsd-schema.adoc
```diff
@@ -0,0 +1,60 @@
...
+
+== Create XML schema for Open Liberty with the Maven plug-in
```
dmuelle: I don't think we need this section. The maven command doesn't have anything to do with the schema, it's just creating a server. While that makes sense for an external blog post, this topic should assume that you already have a server and you want to generate the schema to validate it

modules/reference/pages/xsd-schema.adoc
```diff
@@ -0,0 +1,60 @@
...
+[source,xml]
+----
+java -jar ./wlp/bin/tools/ws-schemagen.jar ./src/main/liberty/config/server.xsd
+----
+
+Set the location of the created XML schema to the same directory that the `server.xml` file exists.
+The default location for the  Open Liberty `server.xml` file is the `src/main/liberty/config` directory.
+
+The command creates XML schema `server.xsd` file in the location that is set in the command.
+
```
dmuelle, line 30: Create XML schema for Open Liberty
--->
Generate an XML schema for your Open Liberty server.xml file

modules/reference/pages/xsd-schema.adoc
```diff
@@ -0,0 +1,60 @@
...
+== Create XML schema for Open Liberty
+
+Download the Open Liberty files and extract the archive, which can be found in the `wlp` directory.
```
dmuelle: Download the Open Liberty files and extract the archive, which can be found in the `wlp` directory.

I'd remove this sentence- the focus is not on creating a server, which is documented elsewhere.

modules/reference/pages/xsd-schema.adoc
```diff
@@ -0,0 +1,60 @@
...
+Create the XML schema by running the following command in the same directory where the `wlp` directory exists:
```
dmuelle: for consistency with other command topics, I'd switch this up to be

Create the XML schema by running the following command in the same directory where the `wlp` directory exists:
---->
Create the XML schema by running the following `java -jar` command from the `_path_to_liberty_/wlp` directory:

modules/reference/pages/xsd-schema.adoc
```diff
@@ -0,0 +1,60 @@
...
+
+[source,xml]
```
dmuelle: since the command is a shell script, not xml, the source should be 

[source,sh]

modules/reference/pages/xsd-schema.adoc
```diff
@@ -0,0 +1,60 @@
...
+----
+java -jar ./wlp/bin/tools/ws-schemagen.jar server.xsd
```
dmuelle: then revise the command:
java -jar ./wlp/bin/tools/ws-schemagen.jar server.xsd
--->
java -jar ./bin/tools/ws-schemagen.jar ./usr/servers/_serverName_/server.xsd

modules/reference/pages/xsd-schema.adoc
```diff
@@ -0,0 +1,60 @@
...
+----
+
+The command creates the XML schema `server.xsd` file in the location that is set in the command.
```
dmuelle: The command creates the XML schema `server.xsd` file in the location that is set in the command.
--->
In this example, the command creates the XML schema `server.xsd` file for the _serverName_ server in the same directory as the server.xml file.

modules/reference/pages/xsd-schema.adoc
```diff
@@ -0,0 +1,60 @@
...
+You can now validate the structure of your `server.xml` file in any IDE of your choice.
```
dmuelle: move this to the end of the next section, as I think you need to tag it in the server.xml in order for the IDE to find it

modules/reference/pages/xsd-schema.adoc
```diff
@@ -0,0 +1,60 @@
...
+After you create the XSD schema, add the schema location to the `xsi : noNamespaceSchemaLocation` attribute in the server element of the `server.xml` file as shown in the following snippet:
+
+[source,xml]
+----
+<server
+ description="my_server"
+ xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+ xsi:noNamespaceSchemaLocation="server.xsd">
+----
+
```
dmuelle, line 56: add clarification:

The schema location is set to `server.xsd` because the example command saved it in the same directory as the server.xml file. If you specify a different path for the output file,  set the noNamespaceSchemaLocation attribute to include the path to the `server.xsd` file

dmuelle (changes requested):
needs to be reworked to be consistent with other command line oriented info in OL docs

dmuelle (approved):

PR #4469, opened 2021-08-03
updates per 4446

dmuelle:
- fix indentations in Gradle snippet
- change source language to Groovy

@lfielke - thanks for the feedback. I'm not sure whether our build will recognize Groovy as a source input but it's worth a shot (my Atom editor doesn't, but I might just need an extension).  In any case, Javascript was wrong so thanks for the catch

### DIFF
--- a/modules/ROOT/pages/relational-database-connections-JDBC.adoc
+++ b/modules/ROOT/pages/relational-database-connections-JDBC.adoc
@@ -67,7 +67,7 @@ configurations {
   jdbcLib
 }
 dependencies {
-	jdbcLib 'com.ibm.db2:jcc:11.5.4.0'
+  jdbcLib 'com.ibm.db2:jcc:11.5.4.0'
 }
 task copyJDBC(type: Copy) {
   from configurations.jdbcLib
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/relational-database-connections-JDBC.adoc
```diff
@@ -61,7 +61,7 @@ If you use Maven to build your application, you can download and deploy the JDBC
 
 If you use Gradle to build your application, you can add a `copy` task to your `build.gradle` file. Then, add a dependency for that task to the `deploy` task in the `build.gradle` file so that the `copy` task runs after the server is created but before the application is deployed:
 
-[source,javascript]
+[source,groovy]
 ----
 configurations {
 	jdbcLib
```
lfielke: nit: reduce indentation of this line and the `jdbcLib` line below to match the new 2-space indent.

dmuelle: fixed

lfielke: Line 70 still has a tab character as the indent, rather than 2 spaces https://github.com/OpenLiberty/docs/pull/4469/files#diff-daefca13d21b4cf1a6e401503fc4d80b065b021efc10615cbcbde1af3ca77dd4R70

lfielke:
LGTM

modules/ROOT/pages/relational-database-connections-JDBC.adoc
dmuelle: thanks- hard to tell from Atom

Rwalls1 (approved):

Issue #3578, opened 2021-02-08
auditUtility command


Charlotte-Holt:
From https://github.com/OpenLiberty/docs/issues/3314, we need to document the `auditUtility` commands. See https://www.ibm.com/support/knowledgecenter/SSEQTP_liberty/com.ibm.websphere.wlp.doc/ae/twlp_audit.html and  https://www.ibm.com/support/knowledgecenter/SSEQTP_liberty/com.ibm.websphere.wlp.doc/ae/rwlp_command_auditutil.html in the KC for more info. This doc should follow the same format as the other command docs.

dmuelle:
need to find exit codes for these commands

emilytee:
Reviewed the draft and approve.

ManasiGandhi:
@dmuelle Can you content review this draft https://draft-openlibertyio.mybluemix.net/docs/21.0.0.9/reference/command/auditUtility-auditReader.html?

dmuelle:
# Content review

Looks good, just a few things to sort. You might compare against the [server package command](https://draft-openlibertyio.mybluemix.net/docs/21.0.0.9/reference/command/server-package.html) for consistency 

- [ ] add to the toc and add a title to the page
- [ ] `The auditUtility auditReader command decrypts and unsigns an audit log that is encrypted and signed.`
The IBM docs source says: `... that is encrypted, signed, or both.`, which is not the same thing. I think IBM docs desc is more precise
- [ ] the current usage example should be where the syntax example is. The usage example should be one of the examples from the "Usage" section of https://www.ibm.com/support/knowledgecenter/SSEQTP_liberty/com.ibm.websphere.wlp.doc/ae/rwlp_command_auditutil.html
- [ ] for consistency with other command topics, remove the actionName section
- [ ] entires for  `auditFileLocation` and `outputFileLocation` should state that they are required options
- [ ] description of `debug` should use complete sentences and fix unclosed monospace on `true`
- [ ] check monospace throughout- `encrypted` description needs monospace around `false`
- [ ] `encKeyStoreLocation` and `signingKeyStoreLocation`- in the description, I think "file URL location" should be "file location". Check with the SME- I see IBM docs says "URL location" but the command help does not. I'm not sure why it would be a URL.  
- [ ] Title for the exit code table is leftover copy/paste

ManasiGandhi:
[...]

PR #4568, opened 2021-08-19
Update auditUtility-auditReader.adoc

ManasiGandhi:
#3578


dmuelle (commented):
the examples look good- for the description, is it clear that the filepaths are examples, and not necessarily what you must configure in all cases?

Since all the options are defined in the table, I wonder if it would be more clear if the example descriptions were simpler, like

`The following example demonstrates the proper syntax to decrypt and audit log that is encrypted:`

similar to what you see in IBM Docs: https://www.ibm.com/docs/en/was-liberty/base?topic=events-auditutility-auditreader-command

dmuelle (approved):

PR #4666, opened 2021-09-20
Update to navigation based on command pages restructuring

ramkumar-k-9286:
Update to navigation based on command pages restructuring.


dmuelle (approved):
LGTM

PR #4667, opened 2021-09-20
Update security vulnerability docs for 21.0.0.10

clarkek123:
This PR is created to update the security vulnerability list with CVEs included in 21.0.0.10.


dmuelle (approved):
LGTM- tested all URLs

Issue #4677, opened 2021-09-21
Document TBASIC and add section for log formats


dmuelle:
A new log format was added in 21.0.0.9, but it was only documented in the release blog. We need to update the [Log and trace configuration page](https://www.openliberty.io/docs/latest/log-trace-configuration.html) to document the new TBASIC format. We should also add a section to this topic that explains the log formats and provides an example. Currently, these formats are not documented in OL and the metatype descriptions are circular and not helpful. 

@frowe opened [an issue to update the metatype](https://github.com/OpenLiberty/open-liberty/issues/18638), we should make sure whatever goes into the new section is consistent with the updated metatype.

### Sources
- [Epic for TBASIC](https://github.com/OpenLiberty/open-liberty/issues/15910) 
- [Release blog post](https://staging-openlibertyio.mybluemix.net/blog/2021/09/03/microprofile-21009.html#tbasic) 
- work with SMEs to get useful descriptions, use cases, and examples for each log format

possible SMEs: @donbourne @fmhwong @tonyreigns

dmuelle:
- [x] Document the TBASIC log format in the[ Configuration settings by source table on the Log and trace configuration page](https://www.openliberty.io/docs/21.0.0.9/log-trace-configuration.html#settings). Add TBASIC as a format option  in the description field for `consoleFormat`, `messageFormat`, and `traceFormat`
- [x] Add a section to the log and trace page for "Log formats". This section should list the available formats, explain why/when a user would choose each format, and provide example output for each.
- [x] Provide links from the table where formats are mentioned to the new log format section.

ManasiGandhi:
Tony's comment in Slack

"The format itself is all identical for all three types of logs, but trace.log won't have the same information. And console.log can sometimes not contain the same information as messages.log, depending on what's configured."

For example
`Messages.log
[20/10/21 16:58:24:515 EDT] 00000001 FrameworkMana A   CWWKE0001I: The server defaultServer has been launched.
[20/10/21 16:58:25:832 EDT] 00000001 FrameworkMana I   CWWKE0002I: The kernel started after 1.491 seconds
[20/10/21 16:58:31:673 EDT] 0000003e FeatureManage A   CWWKF0012I: The server installed the following features: [el-3.0, jsp-2.3, servlet-3.1].
[20/10/21 16:58:31:678 EDT] 0000003e FeatureManage I   CWWKF0008I: Feature update completed in 5.434 seconds.
[...]

PR #4773, opened 2021-10-12
Update log-trace-configuration.adoc

ManasiGandhi:
#4677

### DIFF
--- a/modules/ROOT/pages/log-trace-configuration.adoc
+++ b/modules/ROOT/pages/log-trace-configuration.adoc
@@ -207,11 +207,11 @@ binaryLog view defaultServer
 
 == TBASIC logging format
 
-If you want consistency across your logging formats in Open Liberty you can use the TBASIC logging format.
-A BASIC logging format already exists for `consoleFormat`, `messageFormat`, and `traceFormat`.
+The TBASIC logging format provides a consistent log format across your Open Liberty console log, message log, and trace log files.
+A BASIC logging format exists for `traceFormat`.
 However, the output differs between `traceFormat` and that of the `consoleFormat` and `messageFormat`.
-The new TBASIC format has been added to match the existing BASIC format, providing the ability to have consistent formatting across all three logs.
-The TBASIC format has also been added as a logging format for `traceFormat`, acting as an alias for the existing BASIC option.
+The TBASIC format adds consistency to `consoleFormat`, `messageFormat`, and `traceFormat` to match the BASIC option for `traceFormat`.
+The TBASIC format also acts as an alias for the BASIC option.
 
 The new options can be used in the `bootstrap.properties` file:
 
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/log-trace-configuration.adoc
```diff
@@ -205,6 +205,29 @@ The `binaryLog` command-line tool can be used to convert the binary log to a tex
 binaryLog view defaultServer
 ----
 
+== TBASIC logging format
+
+If you want consistency across your logging formats in Open Liberty you can use the TBASIC logging format.
```
dmuelle: suggested edit-

The TBASIC logging format provides a consistent log format across you Open Liberty console log, message log, and trace log files

modules/ROOT/pages/log-trace-configuration.adoc
```diff
@@ -205,6 +205,29 @@ The `binaryLog` command-line tool can be used to convert the binary log to a tex
...
+A BASIC logging format already exists for `consoleFormat`, `messageFormat`, and `traceFormat`.
```
dmuelle: is this correct? The `logging` autogen doc only lists `BASIC` for the trace log
https://www.openliberty.io/docs/21.0.0.10/reference/config/logging.html

We either need to get dev to edit the autogen or make this info consistent with what is there

I think that BASIC only exists for trace, and TBASIC is added to create a single consistent option for all three logs. But verify with the SME, the autogen might be wrong

modules/ROOT/pages/log-trace-configuration.adoc
```diff
@@ -205,6 +205,29 @@ The `binaryLog` command-line tool can be used to convert the binary log to a tex
...
+However, the output differs between `traceFormat` and that of the `consoleFormat` and `messageFormat`.
+The new TBASIC format has been added to match the existing BASIC format, providing the ability to have consistent formatting across all three logs.
```
dmuelle: avoid saying "new" because it will make the doc out of date when the option is no longer new

also avoid past perfect tense "has been...". Maybe rather than saying it has been added, just say it's avaialble

modules/ROOT/pages/log-trace-configuration.adoc
```diff
@@ -205,6 +205,29 @@ The `binaryLog` command-line tool can be used to convert the binary log to a tex
...
+The TBASIC format has also been added as a logging format for `traceFormat`, acting as an alias for the existing BASIC option.
+
+The new options can be used in the `bootstrap.properties` file:
```
dmuelle, line 218: suggested edit

You can specify the TBASIC log format in your  `bootstrap.properties` file, as shown in the following example:

dmuelle (changes requested):
Need to check technical details against autogen at https://www.openliberty.io/docs/21.0.0.10/reference/config/logging.html and reconcile differences

dmuelle (approved):

Issue #5232, opened 2022-02-23
Add Semeru 17 link to the Java Support table


NottyCode:
The Java Support table on the Java SE Support page doesn't include a link to Semeru 17.

The Java Support table is here: https://openliberty.io/docs/22.0.0.2/java-se.html

Semeru 17 is here: https://developer.ibm.com/languages/java/semeru-runtimes/downloads?os=macOS

ramkumar-k-9286:
Hi David @dmuelle 

The link has been added to the document. The Draft-site link is added below.

https://docs-draft-openlibertyio.mybluemix.net/docs/22.0.0.3/java-se.html

Regards,
Ramkumar

dmuelle:
LGTM- I think you can send it to staging

dmuelle:
please add the java 18 info specified in this PR: https://github.com/OpenLiberty/docs/pull/5250/files

this info should publish with 22.0.0.3 so we can handle it here to avoid branch conflicts

dmuelle:
content is on vNExt and will publish with 22003

PR #5250, opened 2022-03-04
Update java-se.adoc with current supported level of Java 18

jgrassel:
Signed-off-by: Joe Grassel <jgrassel@us.ibm.com>


dmuelle:
we have other changes in this file that have not gone to production yet (but will with 22.0.0.3). Merging these changes will result in branch conflicts. As an alternative, we can create the update as part of #5232 and send the PR for your review

jgrassel:
That sounds good to me.

dmuelle:
actually I just realized these changes have in fact gone to prod. Retargeting this for draft and we will move it to staging and vNext from there.

jgrassel:
BTW, Java 18 support is for 22.0.0.4, not for 22.0.0.3, so make sure that doesn't get into the 20.0.0.3 doc.

PR #5265, opened 2022-03-10
Access log documentation cleanup

mrsaldana:
Cross Component Tracing (XCT) is not supported in Liberty. As such, the %{X}W directive is invalid. The access log documentation should be cleaned up to remove mentions of XCT.


mrsaldana:
addressed by #5266

Issue #5631, opened 2022-07-05
Update `JSON log events reference list` and `Logstash collector events reference list`


Channyboy:
Related EPIC: https://github.com/OpenLiberty/open-liberty/issues/17559

Include the new `ibm_exceptionName` and `ibm_stackTrace` JSON fields to Liberty JSON logging (for messages and trace)

Include the new `exceptionName` and `stackTrace` JSON fields to logstash collector JSON output. (for messages and trace)

dmuelle:
Hi @Channyboy - these updates are now available for review o the draft site:

- [JSON log events reference list: Message events](https://draft-openlibertyio.mybluemix.net/docs/latest/json-log-events-list.html#_message_events)
- [Logstash events reference list: Message events](https://draft-openlibertyio.mybluemix.net/docs/latest/logstash-events-list.html#_message_events)

Please let me know if any further updates are needed. If not, you can add the `Technical reviewed` label to this issue to sign off and the update will publish with the 22.0.0.8 docs release. Thanks!

chirp1:
@dmuelle Hi David, I see "throwable/exception" in the description of various attributes. Would you change the wording so that it doesn't include the slash? I believe the intent is "an exception was thrown", but that wording tends to be a slang- Maybe reword to something like "The type of exception if an exception is logged with the logging event.". Or, starting the description with "if", "If the exception is logged with the logging event, the type of exception..". The description might not need _thrown_.

dmuelle:
@chirp1 - fixed by removing the slash- IBM terminology permits using the term "throwable" as an adjective, eg "a throwable exception"

> throwable  (adjective)
    Usage Use only as an adjective, as in "throwable exception" and "throwable class," and use only in documentation of programming languages that support catch and throw programming blocks, such as Java, C++, Ruby, Lisp, and others.

dmuelle:
changes are on vNext and will publish with 22.0.0.8- thanks!

PR #5632, opened 2022-07-05
5631 json exception fields

Channyboy:
For #5631


dmuelle:
@Channyboy - would you make these commits from a clone rather than a fork? It will make it easier for the docs team to get them to the `staging` and `vNext` branches so they can be published with the proper release

PR #5675, opened 2022-07-26
5637 time based log rollover

dmuelle:


modules/ROOT/pages/access-logging.adoc
```diff
@@ -17,22 +17,22 @@ If you do not specify attributes, the defaults are used. To see a list of the de
 
 == HTTP access log settings
 
-* Using a common log
-+
-To enable logging for multiple endpoints with common settings, include `httpAccessLogging` as a top-level element in your `server.xml` file, and then reference it from multiple `httpEndpoint` elements, as shown in the following example:
+You can configure HTTP access log settings either for multiple endpoints that share common log settings or for individual endpoints.
```
ramkumar-k-9286, line 20:
```suggestion
You can configure HTTP access log settings either for multiple endpoints that share a common log setting or for individual endpoints.
```

dmuelle: I think the plural is approprate in this case, as you can configure many different settings in the `httpAccessLogging` config

ramkumar-k-9286 (approved):

PR #6045, opened 2022-11-18
Remove docker container instuctions and replace with links to Docker Hub

mswatosh:
The docker and podman instructions triple the space required for each database, and will frequently become out of date as new versions are released. By linking to the docker images instead we can save space and users can get the latest images and instructions directly from the image maintainers.

### DIFF
--- a/modules/ROOT/pages/relational-database-connections-JDBC.adoc
+++ b/modules/ROOT/pages/relational-database-connections-JDBC.adoc
@@ -194,30 +194,8 @@ The following example shows a sample data source configuration for a PostgreSQL 
 </dataSource>
 ----
 
-To run a Postgres Docker container locally, run the following command:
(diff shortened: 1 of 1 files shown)


dmuelle:
Thanks for opening this @mswatosh - I think it's a great idea to remove these vendor-specific instructions from our doc and point elsewhere instead.

 I've opened an issue (#6046 ) to connect it to so we can track the changes through to vNext. I'd also like @mbroz2 to have a look just to make sure linking out to Docker Hub is the way to go from a strategic perspective. We've had several issues recently where we've discussed whether to link out to Docker.

mswatosh:
I'm assuming most of those have been pointing to icr.io instead of docker hub? In this case since these are non-IBM (except Db2) they won't be on icr.io, plus docker hub has the included instructions for how to run the images, where icr.io is only image links

PR #6206, opened 2023-02-01
Use short closing form for examples.

NottyCode:
One of the example xml snippets has no child elements, but uses the long form way to close the element. Switching to use the short form.


dmuelle (approved):
LGTM, thanks

PR #6261, opened 2023-02-14
Add MicroProfile 6.0 javadocs to the nav and add a page for it

steven1046:

PR #6272, opened 2023-02-16
Add Microprofile 6.0 to the nav

steven1046:


steven1046:
I accidentally merged this already into the draft branch here: https://github.com/OpenLiberty/docs/pull/6261. Will i need to revert that nav change and is this process still needed?

dmuelle:
The previous merge might cause conflicts the next time draft nav tries to merge into draft. Sorry I missed that in review. Not a big deal though.

For this one, what branch is `Add_MP6_To_Nav` based on? Our process has been to make commits directly to the `draft-nav` branch, which is based on `draft`, then merge that into draft.  If your working branch is based on draft or draft-nav it should be fine?

steven1046:
Add_MP6_To_Nav is branched off of `draft-nav`

steven1046:
Hmm, if we don't revert the previous pr, then this one might not be needed.

dmuelle:
There might be conflicts from the other, but Im not even sure reverting it would prevent them as we can get conflicts just on commit SHA rather than actual content. But they wont be difficult to resolve. We will however duplicate the pattern at staging, rather than using staging-nav to deliver the changes. But simple enough to resolve any conflicts there too.

steven1046:
Thanks, for staging do you think I should put the nav changes into staging-nav?

dmuelle:
I'm not sure :) Maybe just stick with your working branch and we'll sort it out. I dont think it will be a big deal either way.

PR #6289, opened 2023-02-21
Add the MicroProfile 6.0 javadoc

steven1046:
View it on staging: https://docs-staging-openlibertyio.mqj6zf7jocq.us-south.codeengine.appdomain.cloud/docs/latest/reference/javadoc/microprofile-6.0-javadoc.html


dmuelle:
@steven1046 - I'm not sure we're ready to add this to vNExt- MP6 doesn't GA until 23.0.0.3. SOrry- I should have mentioned this in the staging PR. We might need to revert that for the time being

Issue #5814, opened 2022-09-22
Documentation of MicroProfile Metrics


donbourne:
Epic: https://github.com/OpenLiberty/open-liberty/issues/20018

Content: TBD

Things we should document:
1. How to use microprofile metrics with your own micrometer libs so that you can use a different MeterRegistry
    * How to configure meter registries
2. How to use custom scopes for your metrics / how to query for metrics in custom scopes
3. The requirement that the set of tag names for all instances of a metric with the same metric name be the same
----------------------------------
### Existing pages:
* [Monitoring with Metrics](https://openliberty.io/docs/latest/introduction-monitoring-metrics.html) **(Operations > Metrics)** 
   * **Problem: The `MicroProfile Metrics and the /metrics endpoint` section no longer accurately reflect mpMetrics-5.0:**
     * Section makes reference to how to access the `/metrics` endpoint and any sub-trees (i.e. `/metrics/<scope>` and `/metrics/<scope>/<metric name>`. The way to access scope and metric name is no longer with a URL hierarchy, but with parameter query strings (i.e. `/metrics?scope=<scope>` and `/metrics?scope=<scope>&name=<name>`
     * Makes reference to JSON output which is no longer supported.
   * Proposed solution: Where the information is common, leave as is and then separate to a _mpMetrics-4.0 and earlier_ section and `mpMetrics-5.0` section.

* [Metrics reference list](https://openliberty.io/docs/latest/metrics-list.html) ***(Operations > Metrics > Metrics referene list)***
   * **Problem : The existing table is incompatible for mpMetrics-5.0 as well as the few paragraphs above the table.**
      * Prometheus formatted names:
         * scope is no longer prefixed but (i.e. `base_`), but is appended as a tag (i.e. `mp_scope="base"`)
         * Histogram's formatted output no longer contains `stdv` `mean` and `min` values
         * formatting of Prometheus name will always strictly follow `<formattedName>_<unit>_<additionalSuffixes>`
         * `REST.request` metric updated to a `Timer` metric which has a `Histogram` style output. This affects the description as well and the _Version Introduced_.
         * Units are no longer automatically converted.
         * Base and Vendor runtime metrics (except FT related metrics) now register the metric as a base unit. For example, where `cpu.processCpuTime` will show (nanoseconds) in the name column, it will be (seconds) instead.
      * Unrelated to mpMetrics-5.0: Some prometheus names don't show a quotation (`"`) around the tag value.
[...]

PR #6316, opened 2023-02-26
Update to use Influx registry and better explanation on how to retrieve property names

Channyboy:
Use influx as an example. Provide a more intricate explanation on how to resolve the registry and property names

Related to https://github.com/OpenLiberty/docs/issues/5814


modules/ROOT/pages/micrometer-metrics.adoc
```diff
@@ -45,22 +45,28 @@ The following example shows the `server.xml` file configuration for the shared l
...
+  - For example, properties for the Influx registry follow the  `mp.metrics.influx.<property>` syntax. For information about the properties that are available for each registry, see the Micrometer documentation.
+
+Note that the Micrometer documentation for each Micrometer registry will list, as an example, the properties in a format compliant with Spring Boot's ingestion of these properties. The properties follows the syntax `management.metrics.export.<registry>.<property>`. The `<registry>` and `<property>` values correspond to the `<registry>` and `<property>` values used for the above MicroProfile Config properties. Certain properties that are listed may contain dashes. The underlying Micrometer library reads the properties in camel case format. For example, the Micrometer Influx registry documentation lists an `auto-create-db` property that is consumed by Spring Boot. The Micrometer library accepts it as `autoCreateDb`. The corresponding MicroProfile Config property would be `mp.metrics.influx.autoCreateDb`.
```
dmuelle:
```suggestion
The Micrometer documentation for each Micrometer registry lists  example properties in a format that is compliant with the Spring Boot ingestion of these properties. Theese example properties follow the `management.metrics.export.<registry>.<property>` syntax. The `<registry>` and `<property>` values correspond to the `<registry>` and `<property>` values that are used for the previously described MicroProfile Config properties. Certain properties that are listed might contain dashes. The underlying Micrometer library reads the properties in camel case format. For example, the Micrometer Influx registry documentation lists an `auto-create-db` property that is consumed by Spring Boot. The Micrometer library accepts it as `autoCreateDb`. The corresponding MicroProfile Config property is `mp.metrics.influx.autoCreateDb`.
```

PR #6331, opened 2023-03-01
Correct dependencies

Channyboy:
Mistake of assigning micrometer core's dependencies to Prometheus only.
Also left a junit dependency in the pom.xml when generating the dependencies, therefore the junit and hamcrest dependencies are not associated to the meter registries.


Channyboy:
related to https://github.com/OpenLiberty/docs/issues/5814#issuecomment-1446973859

dmuelle (approved):

PR #6452, opened 2023-03-14
3345 fix

### LINKED ISSUE #3345: Suggested security examples
There are a load of security role mapping examples in the JMX KC docs: https://www.ibm.com/support/knowledgecenter/SSAW57_liberty/com.ibm.websphere.wlp.nd.multiplatform.doc/ae/twlp_sec_mapadmin.html

I'm not sure that there's anything specific to JMX about them. Several of them we already have in the OL docs (I think). So I think they should be checked and looked to see whether any of the examples should be used in the OL security docs (eg LDAP role mapping in the LDAP feature gendoc?). The JMX docs (which Manasi has started on in #1695) should then link to the role mapping info in the security topics as appropriate. The JMX docs shouldn't duplicate security info that's elsewhere.

hiltoncv:
#3345 push to staging

### DIFF
--- a/.github/workflows/add-to-proj.yml
+++ b/.github/workflows/add-to-proj.yml
@@ -1,24 +1,19 @@
-name: Add new issues to project
+name: Adds issues with "open" label to GitHub project
 on:
   issues:
-    types: [opened, labeled]
-    
-env:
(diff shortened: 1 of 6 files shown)


dmuelle (approved):
LGTM, thanks

PR #6567, opened 2023-04-04
remove javadoc entries from 23003

dmuelle:


kinueng:
Closing as we no longer need this workaround option to get openliberty.io refreshed with new content.

Issue #5983, opened 2022-11-03
Documentation for Liberty Checkpoint - InstantOn restore server process


dazavala:
For epic https://github.com/OpenLiberty/open-liberty/issues/16384. Opened in accordance to point 6 of [Documenting Open Liberty](https://github.com/OpenLiberty/open-liberty/wiki/Documenting-Open-Liberty).

Create topics in openliberty.io for the new `checkpoint` server command and for InstantOn development.

The discovery for documentation requirements was held Nov11, and openliberty.io requirements are summarized below.  InstantOn also requires [guide documentation](https://github.com/OpenLiberty/guides-common/issues/827) that will publish in openliberty.io. 

Documentation for InstantOn in-container usage may require updates to the DockerHub Open Liberty and IBM Container Repository documentation.  Separate issues will be opened for DockerHub and ICR.

dazavala:
**Summary of openliberty-io user documentation discovery (nov11)**

- [ ] **Basics -> [Container images](https://openliberty.io/docs/latest/container-images.html)**

  - Remove existing documentation for InstantOn container image tags. Once GA all container images will be enabled with the InstantOn capability.
  - Add a subsection to "Build an InstantOn application image" which expands on the previous section "Build an application image".
  - Provide references to the new Basic InstantOn section and Guide 

- [ ] **Basics -> New section for InstantOn**

  - Provide a overview of InstantOn: what is it, why you want it, what you need to use it, and how to use it.
  - What is it: describe the InstantOn capability, mind not to use the term "feature".
  - Why you want it: explain the motivation to scale-to-zero in container. Draw from existing blogs and marketing materials to develop this introduction.
  - What you need to use it OOB: present minimal requirements for Java, OS kernel, CRIU;  describe `depoyment` and `applications` actions for `checkpoint` and explain when to choose one over the other.
  - What you need to use it in container:  explain the limitations in the confines of a container; supported container mgmt tools; list capabilities granted to the container on checkpoint and restore sides; grant operator capabilities (open question!)
  - Document programming models (features) supported by InstantOn (open question!) regarding feature model and potential auto-documentation solutions.

- [ ] **Deployment -> [Open Liberty Operator](https://openliberty.io/docs/latest/open-liberty-operator.html)**

  - Determine whether we must assign operator capabilities from the user.  If so, we may opt to document it here; however, the objective is to address all operator concerns in k8-related documentation (open question!)
  
[...]

PR #6698, opened 2023-06-08
add clarification wrt Ubuntu images

dmuelle:
#5983


dmuelle:
also fix several typos

PR #6730, opened 2023-06-21
Open liberty APIs/SPIs

aknguyen7:

PR #6736, opened 2023-06-22
Revert "Open liberty APIs/SPIs"

aknguyen7:
Reverts OpenLiberty/docs#6730


kinueng:
We are reverting the PR because we need to get a fresh set of files from the build that is related to the next Open Liberty release. The files in the original PR are from previous releases and were from development/testing.

PR #6963, opened 2023-10-11
Update and clarify JWT docs

ayoho:
Add some sections on JWK usage in the JWT builder and consumer doc.

For https://github.com/OpenLiberty/open-liberty/issues/26453

PR #6969, opened 2023-10-12
Add navigation for new JSON Web Tokens page

ayoho:
Follow up to https://github.com/OpenLiberty/docs/pull/6963

### DIFF
--- a/modules/ROOT/nav.adoc
+++ b/modules/ROOT/nav.adoc
@@ -74,6 +74,7 @@
   ** xref:track-loggedout-sso.adoc[Track logged-out SSO cookies]
   ** xref:configuring-spnego-authentication.adoc[Configuring SPNEGO authentication]
   ** xref:enable-openid-connect-client.adoc[Enable an OpenID Connect client]
+* xref:json-web-token.adoc[JSON Web Tokens (JWT)]
 * xref:authorization.adoc[Authorization]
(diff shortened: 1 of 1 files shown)


modules/ROOT/nav.adoc
```diff
@@ -74,6 +74,7 @@
   ** xref:track-loggedout-sso.adoc[Track logged-out SSO cookies]
   ** xref:configuring-spnego-authentication.adoc[Configuring SPNEGO authentication]
   ** xref:enable-openid-connect-client.adoc[Enable an OpenID Connect client]
+* xref:json-web-token.adoc[JSON Web Tokens (JWT)]
```
dmuelle, line 77: wondering if this should be nested in the "Single Sign-On" section?

dmuelle, line 77:
```suggestion
    ** xref:json-web-token.adoc[JSON Web Tokens (JWT)]
```

ayoho: I didn't want it to be nested under Single Sign-On since I don't really consider JWTs to be an SSO technology. I'd be fine nesting it if you had a stronger opinion on it, though.

dmuelle: interesting- ok we can keep it as you have it then

dmuelle (approved):
LGTM

Issue #6985, opened 2023-10-19
Add missing Liberty container image tag and reorder the tags


leochr:
Document: Open Liberty container images
http://openliberty.io/docs/latest/container-images.html

**Please describe the problem you are having with the documentation. Is information missing, inaccurate, or unclear? Tell us about the context where you encountered the problem so we can understand how to address it.**

`kernel-slim-java8-ibmjava-ubi` is missing from the list of images in "[Container images on IBM Container Registry](https://openliberty.io/docs/latest/container-images.html#_container_images_on_ibm_container_registry)"

Also, please reorder the tags to list the newer Java versions at the top. Like this:

```
kernel-slim-java17-openj9-ubi
kernel-slim-java11-openj9-ubi
kernel-slim-java8-openj9-ubi
kernel-slim-java8-ibmjava-ubi
full-java17-openj9-ubi
full-java11-openj9-ubi
full-java8-openj9-ubi
full-java8-ibmjava-ubi
beta
latest
```

ramkumar-k-9286:
Hi @leochr 

The suggested correction was done.

Draft link:
https://docs-draft-openlibertyio.mqj6zf7jocq.us-south.codeengine.appdomain.cloud/docs/latest/container-images.html

Please review and add the `Developer Reviewed` label if you are satisfied with the changes.

Regards,
Ramkumar.

CC @dmuelle

leochr:
@ramkumar-k-9286 Thanks for the updates. Looks good.

dmuelle:
LGTM, thanks. Send to staging when you have a chance

ramkumar-k-9286:
@dmuelle 

Hi David

Staging page link:

https://docs-staging-openlibertyio.mqj6zf7jocq.us-south.codeengine.appdomain.cloud/docs/latest/container-images.html

Regards,
Ramkumar.

PR #7013, opened 2023-11-01
6985-container image tag and reorder the tags-1

ramkumar-k-9286:
container image tag and reorder the tags-1

#6985


dmuelle (approved):
LGTM

PR #7049, opened 2023-11-12
Mention <executor> element in thread pool tuning

scottkurz:
It seems just a bit incomplete to mention the need to set the two specific attributes, but without any mention of the element of which these are attributes.

Something like this I think is all that's needed, (however worded).

True, there's a "See also" link to the executor elem config at the bottom, but maybe it should appear in this article.

Or, we could go to the lengths of adding a sample element w/ attrs.


dmuelle:
superseded by #7064

PR #7064, opened 2023-11-17
Minor edits

dmuelle:
fix typos and clarify executor config per #7061 and #7049


modules/ROOT/pages/thread-pool-tuning.adoc
```diff
@@ -91,6 +92,8 @@ Some environments set a hard limit on the number of threads that a process can c
 Currently, Open Liberty has no way to know whether such a cap applies, or what the value is.
 If Open Liberty is running in a thread-limited environment, the operator can configure the `maxThreads` attribute to an acceptable value.
 
+
```
dmuelle, line 95: Hi @scottkurz - we could add a quick config example here for core/max threads- do you know what realistic example values for these attributes would be?

dmuelle:
Hi @scottkurz - thanks for your recent edits- consolidating them to one PR here for convenience. I made slight edits to the thread pool info, let me know what you think.

modules/ROOT/pages/thread-pool-tuning.adoc
scottkurz: Good question.  I skimmed https://openliberty.io/blog/2019/04/03/liberty-threadpool-autotuning.html and I'm not sure.  We could ask @gjdeval  I guess.  Or does not including an example align with the implicit point here... if you think you need to set this well, there's a good chance you're wrong, so don't make it any easier.  IDK.  

Otherwise the changes look good.

scottkurz (approved):

modules/ROOT/pages/thread-pool-tuning.adoc
dmuelle: Yes- I seem to recall that might have been Gary's take when he wrote that post, which was the source for this doc. I'll leave the example out for now and we ca always add later if there's a need

gjdeval: The changes in the doc look fine. We could refer them to the blog post on the subject for more guidance, if that is the sort of thing we do in our "official" doc.

https://openliberty.io/blog/2019/04/03/liberty-threadpool-autotuning.html

Issue #7090, opened 2023-11-29
Documentation on OCP supported versions for InstantOn


tjwatson:
Not sure if we should put this directly in the InstantOn documentation or into the general Liberty documentation for OCP support.

tjwatson:
@tam512 could you provide the instructions you use to setup the necessary Linux capabilities with OCP 4.14 so we can add that to the InstantOn documentation?

tam512:
First I use the following steps provided by @anjumfatima90 to create service account and security context constraints

```
1. Create a service account from command line by running following command.  Note that  you can use any serviceaccount name instead of instanton-sa, but you need to refer to that name in the later steps.
    a. oc create serviceaccount instanton-sa

2. Add the security constraints to the service account.
    a. Create a file scc-cap-cr.yaml. Add the following in the file

allowHostDirVolumePlugin: true
allowHostIPC: false
allowHostNetwork: false
allowHostPID: false
allowHostPorts: false
allowPrivilegeEscalation: true
allowPrivilegedContainer: false
allowedCapabilities: null
apiVersion: security.openshift.io/v1
defaultAddCapabilities:
- CHECKPOINT_RESTORE
- SETPCAP
fsGroup:
  type: MustRunAs
groups:
- system:authenticated
kind: SecurityContextConstraints
metadata:
  annotations:
    kubernetes.io/description: criu-scc is based on the restricted SCC but removes any restrictions
      that prevent the restore image from running successfully.
  generation: 1
  name: cap-cr-scc
priority: null
readOnlyRootFilesystem: false
requiredDropCapabilities:
- KILL
- MKNOD
- SETUID
- SETGID
runAsUser:
  type: RunAsAny
  #uid: 1001
seLinuxContext:
  type: MustRunAs
supplementalGroups:
  type: RunAsAny
users: []
volumes:
- configMap
- downwardAPI
- emptyDir
- persistentVolumeClaim
- projected
- secret


3. Apply the scc to the service account by running following commands
    a. oc apply -f scc-cap-cr.yaml
    b. oc adm policy add-scc-to-user cap-cr-scc -z instanton-sa
    Note: In the above command the name of the yaml file doesn't matter, it's the name specified inside the file.
```

[...]

PR #7119, opened 2023-12-07
add OCP SCC info

dmuelle:
#7090

### DIFF
--- a/modules/ROOT/pages/instanton.adoc
+++ b/modules/ROOT/pages/instanton.adoc
@@ -338,6 +338,36 @@ When you deploy to Kubernetes, the container must be granted the `CHECKPOINT_RES
             - ALL
 ----
 
+==== Red Hat OpenShift security context constraints
+
+To deploy applications to Red Hat OpenShift with InstantOn, you must specify a security context constraint (SCC) that at a minimum grants the capabilities to the defaults that are needed for InstantOn. The following example defines an SCC with these capabilities that is called `defaultAddCapabilities`:
+
+[source,yaml]
+----
+defaultAddCapabilities:
+- CHECKPOINT_RESTORE
+- SETPCAP
+----
+
+The applications you deploy must be associated with this SCC, for example, by specifying the `serviceAccountName` setting in the deployment yaml file. The following example specifies the `defaultAddCapabilities` SCC that was defined in the previous example:
+
+[source,yaml]
+----
+  serviceAccountName: defaultAddCapabilities
+  securityContext:
+    allowPrivilegeEscalation: true
+    privileged: false
+    runAsNonRoot: true
+    capabilities:
+      add:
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/instanton.adoc
```diff
@@ -338,6 +338,36 @@ When you deploy to Kubernetes, the container must be granted the `CHECKPOINT_RES
             - ALL
 ----
 
+==== Red Hat OpenShift security context constraints
+
+To deploy applications to Red Hat OpenShift with InstantOn, you must specify a security context constraint (SCC) that at a minimum grants the capabilities to the defaults that are needed for InstantOn. The following example defines an SCC with these capabilities that is called `defaultAddCapabilities`:
```
tjwatson, line 343: This doesn't read right to me.
> that at a minimum grants the capabilities to the defaults that are needed for InstantOn.

The capabilities listed under `defaultAddCapabilities` are adding the capabilities to the set of capabilities which are granted to any pod.  So it is adding capabilities to that list of default capabilities.  To me it would make more sense to read like this:

> that at a minimum specifies a list of additional capabilities that are added to any pod. The following example defines an SCC with the required capabilities by using the parameter called `defaultAddCapabilities`:

modules/ROOT/pages/instanton.adoc
```diff
@@ -338,6 +338,36 @@ When you deploy to Kubernetes, the container must be granted the `CHECKPOINT_RES
...
+
+[source,yaml]
+----
+defaultAddCapabilities:
+- CHECKPOINT_RESTORE
+- SETPCAP
+----
+
+The applications you deploy must be associated with this SCC, for example, by specifying the `serviceAccountName` setting in the deployment yaml file. The following example specifies the `defaultAddCapabilities` SCC that was defined in the previous example:
```
tjwatson, line 352: This is confusing because `defaultAddCapabilities` is treated as if it is an SCC name, but before `defaultAddCapabilities` is a parameter of the SCC yaml file.  I think it would be better to use a name like `liberty-instanton-scc`:

> The applications you deploy must be associated with an SCC that adds the required capabilities. For example, assume there is an SCC deployed called `liberty-instanton-scc` that added the required capabilities.  The following example specifies the `serviceAccountName` parameter to set the SCC name to `liberty-instanton-scc`:

modules/ROOT/pages/instanton.adoc
```diff
@@ -338,6 +338,36 @@ When you deploy to Kubernetes, the container must be granted the `CHECKPOINT_RES
...
+
+[source,yaml]
+----
+  serviceAccountName: defaultAddCapabilities
```
tjwatson, line 356: Change to:
```
  serviceAccountName: liberty-instanton-scc
```

PR #7307, opened 2024-04-29
Add section Considerations when using a proxy server

### LINKED ISSUE #7223: Document wsAtomicTransaction externalURLPrefix limitations in Kubernetes
When using [`wsAtomicTransaction`](https://openliberty.io/docs/latest/reference/config/wsAtomicTransaction.html) in a Kubernetes environment without [peer recovery](https://openliberty.io/docs/latest/transaction-service.html#cloud), if `externalURLPrefix` is set to use `${env.POD_IP}`, and if the client pod communicates through a Kubernetes service, and if the service communication is NATted to the pod IP (e.g. OpenShift SDN), then, under heavy load, if a socket opened to the service IP of the target pod is closed and enters `TIME_WAIT`, and if a new socket is attempted to be created by WS-AT on the pod IP and it happens to use the same ephemeral local port, then `SYN` packets will be dropped and retransmitted during the `TIME_WAIT` period causing undesired latency. If this duration exceeds 30 seconds, the WS-AT call will fail with a `java.net.SocketTimeoutException` with diagnostic trace `*=info:org.apache.cxf.*=all`:

```
java.net.SocketTimeoutException: connect timed out
        [...]
        at com.ibm.ws.wsat.service.impl.WebClientImpl$3.call(WebClientImpl.java:120)
```

This manifests as retransmitting `SYN` packets in network trace on the client; for example:

```
$ TZ=UTC tshark -t ud -T fields -e frame.number -e _ws.col.Time -e ip.src -e tcp.srcport -e ip.dst -e tcp.dstport -e tcp.stream -e frame.len -e _ws.col.Protocol -e _ws.col.Info -r *pcap* -Y "tcp.flags.syn == 1 && tcp.analysis.retransmission"
409708	2024-02-05 16:59:08.192371	10.1.2.3	5502	10.1.2.4	9443	3946	76	TCP	[TCP Retransmission] 5502 → 9443 [SYN] Seq=0 Win=26733 Len=0 MSS=8911 SACK_PERM=1 TSval=123 TSecr=0 WS=128
411507	2024-02-05 16:59:10.241377	10.1.2.3	5502	10.1.2.4	9443	3946	76	TCP	[TCP Retransmission] 5502 → 9443 [SYN] Seq=0 Win=26733 Len=0 MSS=8911 SACK_PERM=1 TSval=123 TSecr=0 WS=128
412178	2024-02-05 16:59:14.272379	10.1.2.3	5502	10.1.2.4	9443	3946	76	TCP	[TCP Retransmission] 5502 → 9443 [SYN] Seq=0 Win=26733 Len=0 MSS=8911 SACK_PERM=1 TSval=123 TSecr=0 WS=128
419229	2024-02-05 16:59:22.784377	10.1.2.3	5502	10.1.2.4	9443	3946	76	TCP	[TCP Retransmission] 5502 → 9443 [SYN] Seq=0 Win=26733 Len=0 MSS=8911 SACK_PERM=1 TSval=123 TSecr=0 WS=128
```

In other words, because of the NATting, although the client sends to the service IP, the pod receives the traffic on the pod IP, so although the client's 4-tuple is (client IP, client ephemeral port, service IP, service port), the pod's 4-tuple is (client IP, client ephemeral port, pod IP, pod port). Example tcpdumps demonstrating this differing destination IP (the 4-tuple is columns 3-6):

Client pod:
```
389156	2024-02-05 16:58:30.554889	10.1.2.3	5502	172.1.2.3	9443	3709	68	TCP	5502 → 9443 [ACK] Seq=4697 Ack=635 Win=123 Len=0 TSval=123 TSecr=123
```

Target pod:
```
554314	2024-02-05 16:58:30.554905	10.1.2.3	5502	10.1.2.4	9443	11791	68	TCP	5502 → 9443 [ACK] Seq=4697 Ack=635 Win=123 Len=0 TSval=123 TSecr=123
```

So when the socket to the service is closed and enters `TIME_WAIT`, if WS-AT then sends a new socket directly to the pod IP (client IP, client ephemeral port, pod IP, pod port), the pod's 4-tuple is the same as before (client IP, client ephemeral port, pod IP, pod port) and cannot be re-used. It is believed this is due to the default [conntrack `TIME_WAIT` time of 120 seconds](https://www.kernel.org/doc/Documentation/networking/nf_conntrack-sysctl.txt). Using [`net.ipv4.tcp_tw_reuse=1`](https://www.kernel.org/doc/Documentation/networking/ip-sysctl.txt) does not help. It is possible that [`net.netfilter.nf_conntrack_tcp_timeout_time_wait=60`](https://www.kernel.org/doc/Documentation/networking/nf_conntrack-sysctl.txt) may resolve the issue.

One workaround is to use a different port for `externalURLPrefix` than what is used for the service which avoids the 4-tuple TCP conflict. This requires adding an additional `httpEndpoint` and adding the port `hostAlias` to the `virtualHost`. Since a `virtualHost` may not have been explicitly defined previously and the default was used, take care to include all incoming virtual hosts. In a Kubernetes environment, although traffic might come into a particular port (e.g. 9443), the request `Host` header might be on a service port (e.g. 443) and missing such a `hostAlias` will result in 404s. For example:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<server>
  <httpEndpoint id="wsatHttpEndpoint" host="*" httpsPort="9444" />
  <wsAtomicTransaction SSLEnabled="true" SSLRef="cssSSLSettings" externalURLPrefix="https://${env.POD_IP}:9444" />
  <virtualHost id="default_host">
    <hostAlias>*:443</hostAlias>
    <hostAlias>*:9443</hostAlias>
    <hostAlias>*:9444</hostAlias>
  </virtualHost>
</server>
```

In addition, ensure that the new port is accessible by the client. For example, although a `ports.containerPort` entry on the pod specification is [not needed](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#ports) as "Any port which is listening on the default '0.0.0.0' address inside a container will be accessible from the network", there may be a [NetworkPolicy](https://kubernetes.io/docs/concepts/services-networking/network-policies/) that restricts available ports and thus an additional `NetworkPolicy` may be required.

While this issue affects [OpenShift SDN networking](https://docs.openshift.com/container-platform/latest/networking/openshift_sdn/about-openshift-sdn.html) it does not appear to affect [OVN-Kubernetes networking](https://docs.openshift.com/container-platform/latest/networking/ovn_kubernetes_network_provider/about-ovn-kubernetes.html). OVN-Kubernetes [has replaced](https://docs.openshift.com/container-platform/4.12/release_notes/ocp-4-12-release-notes.html#ocp-4-12-ovn-kubernetes-default-network-plug-in) OpenShift SDN as the default networking plugin as of OpenShift 4.12. It is believed that switching to OVN-Kubernetes may also resolve this issue.

Diagnostic notes:

1. Gather `tcpdump` on both client and server pods using `nsenter`: https://access.redhat.com/solutions/4569211 and https://access.redhat.com/solutions/1611883
1. Quickly finding retransmitting SYNs on the socket connect:
   \
   `TZ=UTC tshark -t ud -T fields -e frame.number -e _ws.col.Time -e ip.src -e tcp.srcport -e ip.dst -e tcp.dstport -e tcp.stream -e frame.len -e _ws.col.Protocol -e _ws.col.Info -r *pcap* -Y "tcp.flags.syn == 1 && tcp.analysis.retransmission"`
   
    1. Then, search for that same 4-tuple before the first `SYN` and check if any packets (usually `FIN` or `ACK`, but could also be `RST`) are within [60 seconds](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/net/tcp.h?h=v6.7#n124) of the first `SYN`.

kgibm:
Fixes #7223

### DIFF
--- a/modules/ROOT/pages/transaction-service.adoc
+++ b/modules/ROOT/pages/transaction-service.adoc
@@ -85,6 +85,19 @@ To monitor peers, servers must be configured so that their transaction logs are 
 
 The use of a shared file system for this purpose, such as an RWX persistent volume in a Kubernetes cluster, is explicitly not supported where the file system crosses data centers. This limitation is due to the difficulty of implementing POSIX locking semantics in such a configuration. Under these circumstances, the transaction service must be configured to use a relational database for its transaction logs. For more information, see the <<#log,Transaction log management>> section.
 
+=== Considerations when using a proxy server
+
+The older alternative to peer recovery in a cloud environment is to route WS-AtomicTransaction traffic through a proxy server using `externalURLPrefix`. This may be done by setting `externalURLPrefix` to the pod IP address; for example:
+
+[source,xml]
+----
+<wsAtomicTransaction
+  ...
+  externalURLPrefix="https://${env.POD_IP}:9443"
+/>
+----
+
+It has been observed in some Kubernetes environments that this may cause intermittent socket timeout exceptions when non-WS-AtomicTransaction traffic communicates through a Kubernetes service to the same destination pod on the same port. To avoid this, consider adding an additional config:httpEndpoint[display=HTTP Endpoint] with a different port solely for `externalURLPrefix` traffic, add an additional `hostAlias` for this port in the config:virtualHost[display=Virtual Host], and configure `externalURLPrefix` to use this new port.
 
 [#log]
 == Transaction log management
--- a/modules/ROOT/pages/transaction-service.adoc
+++ b/modules/ROOT/pages/transaction-service.adoc
@@ -87,7 +87,7 @@ The use of a shared file system for this purpose, such as an RWX persistent volu
 
 === Considerations when using a proxy server
 
-The older alternative to peer recovery in a cloud environment is to route WS-AtomicTransaction traffic through a proxy server using `externalURLPrefix`. This may be done by setting `externalURLPrefix` to the pod IP address; for example:
+The older alternative to peer recovery in a cloud environment is to route WS-AtomicTransaction traffic through a proxy server by specifying the `externalURLPrefix` attribute in the `server.xml` file. Set the attribute` to the pod IP address, as shown in the following in the example:
 
 [source,xml]
 ----
@@ -97,7 +97,7 @@ The older alternative to peer recovery in a cloud environment is to route WS-Ato
 />
 ----
 
-It has been observed in some Kubernetes environments that this may cause intermittent socket timeout exceptions when non-WS-AtomicTransaction traffic communicates through a Kubernetes service to the same destination pod on the same port. To avoid this, consider adding an additional config:httpEndpoint[display=HTTP Endpoint] with a different port solely for `externalURLPrefix` traffic, add an additional `hostAlias` for this port in the config:virtualHost[display=Virtual Host], and configure `externalURLPrefix` to use this new port.
+In some Kubernetes environments, this setting might cause intermittent socket timeout exceptions when non-WS-AtomicTransaction traffic communicates through a Kubernetes service to the same destination pod on the same port. To avoid these timeouts, add another config:httpEndpoint[display=HTTP Endpoint] with a different port that is solely for `externalURLPrefix` traffic. Then, add another `hostAlias` attribute for this port in the config:virtualHost[display=Virtual Host] configuration and set the `externalURLPrefix` attribute to this new port.
 
 [#log]
 == Transaction log management
--- a/modules/ROOT/pages/transaction-service.adoc
+++ b/modules/ROOT/pages/transaction-service.adoc
@@ -87,7 +87,7 @@ The use of a shared file system for this purpose, such as an RWX persistent volu
 
 === Considerations when using a proxy server
 
-The older alternative to peer recovery in a cloud environment is to route WS-AtomicTransaction traffic through a proxy server by specifying the `externalURLPrefix` attribute in the `server.xml` file. Set the attribute to the pod IP address, as shown in the following in the example:
+The older alternative to peer recovery in a cloud environment is to route WS-AtomicTransaction traffic through a proxy server by specifying the `externalURLPrefix` attribute in the `server.xml` file by setting the attribute to the pod IP address, as shown in the following in the example:
 
 [source,xml]
 ----
--- a/modules/ROOT/pages/transaction-service.adoc
+++ b/modules/ROOT/pages/transaction-service.adoc
@@ -87,7 +87,7 @@ The use of a shared file system for this purpose, such as an RWX persistent volu
 
 === Considerations when using a proxy server
 
-The older alternative to peer recovery in a cloud environment is to route WS-AtomicTransaction traffic through a proxy server by specifying the `externalURLPrefix` attribute in the `server.xml` file by setting the attribute to the pod IP address, as shown in the following in the example:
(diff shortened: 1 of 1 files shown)


dmuelle:
Hi @kgibm - Thanks for this update. Updates to the docs need to target `draft` and then `staging` before they can go to vNext. I'll change the base to draft here. The update process is documented here: https://github.com/OpenLiberty/docs?tab=readme-ov-file#updating-an-existing-topic. Although this process says to clone the repo, you can also commit from your fork.

djmatthews (approved):

modules/ROOT/pages/transaction-service.adoc
```diff
@@ -85,6 +85,19 @@ To monitor peers, servers must be configured so that their transaction logs are
 
 The use of a shared file system for this purpose, such as an RWX persistent volume in a Kubernetes cluster, is explicitly not supported where the file system crosses data centers. This limitation is due to the difficulty of implementing POSIX locking semantics in such a configuration. Under these circumstances, the transaction service must be configured to use a relational database for its transaction logs. For more information, see the <<#log,Transaction log management>> section.
 
+=== Considerations when using a proxy server
+
+The older alternative to peer recovery in a cloud environment is to route WS-AtomicTransaction traffic through a proxy server using `externalURLPrefix`. This may be done by setting `externalURLPrefix` to the pod IP address; for example:
```
dmuelle:
```suggestion
The older alternative to peer recovery in a cloud environment is to route WS-AtomicTransaction traffic through a proxy server by specifying the `externalURLPrefix` attribute in the `server.xml` file. Set the attribute` to the pod IP address, as shown in the following in the example:
```

modules/ROOT/pages/transaction-service.adoc
```diff
@@ -85,6 +85,19 @@ To monitor peers, servers must be configured so that their transaction logs are
...
+
+[source,xml]
+----
+<wsAtomicTransaction
+  ...
+  externalURLPrefix="https://${env.POD_IP}:9443"
+/>
+----
+
+It has been observed in some Kubernetes environments that this may cause intermittent socket timeout exceptions when non-WS-AtomicTransaction traffic communicates through a Kubernetes service to the same destination pod on the same port. To avoid this, consider adding an additional config:httpEndpoint[display=HTTP Endpoint] with a different port solely for `externalURLPrefix` traffic, add an additional `hostAlias` for this port in the config:virtualHost[display=Virtual Host], and configure `externalURLPrefix` to use this new port.
```
dmuelle:
```suggestion
In some Kubernetes environments, this setting might cause intermittent socket timeout exceptions when non-WS-AtomicTransaction traffic communicates through a Kubernetes service to the same destination pod on the same port. To avoid these timeouts, add another config:httpEndpoint[display=HTTP Endpoint] with a different port that is solely for `externalURLPrefix` traffic. Then, add another `hostAlias` attribute for this port in the config:virtualHost[display=Virtual Host] configuration and set the `externalURLPrefix` attribute to this new port.
```

dmuelle (changes requested):
ID review, minor edits per IBM style

dmuelle (approved):
LGTM thanks

Issue #7399, opened 2024-06-24
Update Documentation for MicroProfile Metrics 3.0/4.0


tonyreigns:
Related to https://github.com/OpenLiberty/open-liberty/issues/25519

MP Metrics 3.0/4.0 will include some backported MP Config properties that were introduced in Metrics 5.1, see https://github.com/OpenLiberty/docs/issues/6792

The properties are identical. The only updates being added are mentioning that the properties are available starting from Metrics 3.0 rather than 5.1.

Updates would be needed here:

https://openliberty.io/docs/latest/microprofile-config-properties.html#metrics 
https://openliberty.io/docs/latest/microservice-observability-metrics.html

PR #7400, opened 2024-06-24
Documentation for theMP Config properties backport for MP Metrics 3.0 and 40

tonyreigns:
#7399 

Just looking for approval from Metrics SME before requesting additional review the docs team. Thanks!


Channyboy:
@dmuelle  this is back-porting functionality to earlier versions of `mpMetrics`. Not sure if we need to say "as of 24.0.0.x" you can start using this in `mpMetrics-3.0` and above.  Or if we just say it outright like that.

dmuelle:
@Channyboy - since each version of the docs is hosted in it's own branch (selectable via the version picker), this content will only show in versions 24007 and later, so no need to call out the version.

dmuelle (approved):
LGTM

